PR #4, opened 2015-11-16
updated pip packaging

shivarammysore:


KitL:
The only issue I have with this stuff is in the deployment picture you mention a northbound api, which doesnt exist yet. Nor am I sure to what extent it will exist. I'd be happier if that wasnt there. I'm going to merge the other commit though..

Also try and be detailed in your commit messages. Often they are fairly self-explanatory, but a little detail about why you are making a change is always nice.

KitL:
Ok damn, I also failed to notice that you have reserved all rights when this should be apache licensed.

I'll get onto that.

PR #64, opened 2016-04-20
Some gauge bug fixes and documentation updates

rsanger:
Fixes gauge logging and logging documentation
Fixes indent bug causing a crash when a datapath leaves
Fixes a typo bug when not using influxdb


anarkiwi:
Some gauge bug fixes and documentation updates.

PR #72, opened 2016-04-27
Add fine-grained port mirroring and associated tests 

gwacter:
These are the changes associated with the work done at the Hackfest in Sydney (2016). Have tested against existing tests and added two additional tests for new ACL format and ACL mirror.

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/valve.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/valve.py
@@ -548,8 +548,14 @@ class OVSStatelessValve(Valve):
                 acl_inst = []
                 match_dict = {}
                 for attrib, attrib_value in rule_conf.iteritems():
-                    if attrib == 'allow':
-                        if attrib_value == 1:
+                    if attrib == "actions":
+                        if 'mirror' in attrib_value:
+                            port_no = attrib_value['mirror']
+                            acl_inst.append(
+                                    self.apply_actions([
+                                    #parser.OFPActionPopVlan(),
+                                    parser.OFPActionOutput(port_no)]))
+                        if attrib_value['allow'] == 1:
                             acl_inst.append(acl_allow_inst)
                         continue
                     if attrib == 'in_port':
@@ -697,6 +703,9 @@ class OVSStatelessValve(Valve):
         self.logger.info('Sending config for port {0}'.format(port))
 
         for table in self.all_valve_tables():
+            if table == self.dp.acl_table:
+                # Shouldn't delete flows from acl_table
+                continue
             ofmsgs.append(self.valve_flowdel(table, in_port_match))
 
         if port_num in self.dp.mirror_from_port.values():
@@ -733,6 +742,8 @@ class OVSStatelessValve(Valve):
 
         # delete all rules matching this port in all tables.
         for table in self.all_valve_tables():
+            if table == self.dp.acl_table:
+                continue
             ofmsgs.append(self.valve_flowdel(table,
                 self.valve_in_match(in_port=port_num)))
 
--- a/src/ryu_faucet/org/onfsdn/faucet/valve.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/valve.py
@@ -544,7 +544,6 @@ class OVSStatelessValve(Valve):
             acl_rule_priority = self.dp.highest_priority
             acl_allow_inst = self.goto_table(self.dp.eth_src_table)
             for rule_conf in self.dp.acls[acl_num]:
-                # default drop
                 acl_inst = []
                 match_dict = {}
                 for attrib, attrib_value in rule_conf.iteritems():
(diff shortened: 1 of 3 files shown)


anarkiwi:
Git reports merge conflicts.

Please resolve these conflicts and we can review again.

src/ryu_faucet/org/onfsdn/faucet/valve.py
```diff
@@ -618,6 +627,8 @@ def port_delete(self, dp_id, port_num):
 
         # delete all rules matching this port in all tables.
         for table in self.all_valve_tables():
+            #if table == self.dp.acl_table:
+	    #  continue
```
KitL: This shouldn't be commented right?

gwacter: You are right -- it shouldn't be commented. We don't want to drop ACL rules unless the user requests the change.

gwacter:
Conflicts removed, passing all tests except for MAX HOSTS.

anarkiwi:
Great, thanks, looks better.

There are other changes to the base test class. Why are they necessary?

There is a magic constant: self.mirror_from_port[4294967040 - acl_num] =
port_no. Where does this number come from and why is it necessary?

This change attempts to maintain ACL rules between calls. I'm not opposed
to this in principle but let's not try to do this in this particular change
- we should focus solely on the new functionality added. Please remove that
code from this change.

anarkiwi:
Thanks also for reporting the flakeyness in the MaxHosts test (it
occasionally fails in my setup also - OVS 2.4.1). I just committed a change
to the test to make it retry, if you'd like to merge that.

anarkiwi:
Ping - is it possible to get these changes actioned please? With the requested changes the diff should be very small and small changes will be easy to verify and merge.

Callum027:
Sorry for the late response, was not subscribed to this thread at first.

All tests seem to be passing on my end - @trungdtbk, could you please verify that everything is fine?

trungdtbk:
Hi Callum,
I run the tests. Maxhosts and Mirrortest fail if I run all the tests together. But they are successful if run individually.

anarkiwi:
Hi all;

Let's reset.

Trung - please can you see the comments earlier in this thread, about the
changes to the tests. We want to trim your change down to the minimal one
(I see for example, that you hard code OF1.3 in the test base class - why
is this is necessary)?

If there are changes needed to the base class, please add notes explaining
why to the pull request.

Thanks,

trungdtbk:
Hi Josh,
I used mininet 2.2.0 and it fails to create OF1.3 switches without specifying 'OpenFlow13'. 

Regarding "4294967040" (=MAX of OF ports), I'm not so sure why we use that. But may be for avoiding collision because we use the same dict for both port mirroring and acl mirroring.

I've removed the codes at 707-708 regarding ACL rules.
Cheers,

anarkiwi:
OK. We need to figure out why you need OF1.3 to be specified. It isn't
required in any other test scenario I know of. Which version of OVS are you
testing with?

We can't accept a change if we're not sure what the code does. :) What do
you intend to happen?

trungdtbk:
I use ovs 2.0.2

Suppose if we remove the magic number, then the following config will create anomaly.
Port 1 has its traffic mirrored to port 2 by acl rule
Port 3 also has its traffic mirrored to port 2 (unwanted result)
interfaces:
     1:
            acl_in: 3
      2:
      3:
acls:
     3:
           -rule:
                  actions:
                         allow: 1
                         mirror: 2

anarkiwi:
Per the comments in the test file, OVS 2.0.2 is unsupported.

Please use the recommend version and repeat the test.

gwacter:
@trungdtbk  and I had a chat. We think that if we use the right version of OVS that we can remove the Openflow13 flag and return the print statements in faucet test back to the previous version (they were changed to work with the older version of OVS).

Regarding the magic number. It is a bit of hack to allow reuse of the mirror_from_port data structure belonging to dp so it can record both the case where all traffic is mirrored and the selective case. It works, might need refactoring in the longer term.

anarkiwi:
Great (re OVS).

I'm sorry, we can't have the magic number hack. I'm happy to hear another
proposal (or discuss in person on Wednesday). Now that we are at V1.0 and
more and more people are relying on this code, we can't commit experimental
code to the codebase.

gwacter:
I propose that we add an extra data structure to dp.py to do the housekeeping for fine grained mirroring. I think the only other change would be to modify the function **add_ports_and_vlans** to add any ports that are the target of those acls to the list of ports to be configured. 

@trungdtbk is going to work on that tomorrow.

anarkiwi:
That sounds a lot cleaner. Great!

gwacter:
@trungdtbk has simplified the code (removed extra data structure and optimisations re multiple drop flows).

gwacter:
@trungdtbk just confirmed that it passes all the tests.

anarkiwi:
I have merged it. This is a useful feature! I look forward to seeing it
used.

PR #88, opened 2016-05-24
Add Dockerfiles for running locally-built faucet, and running mininet tests

simeonmiteff:
This adds a new make target and some new docker containers:

- **Dockerfile.dev** uses the ryu-faucet package built by the new `make dockerdev` target instead of installing the current package with pip.

- **Dockerfile.tests** does the same thing as Dockerfile.dev, but runs the mininet tests as its entrypoint (it builds mininet and openvswitch, and other packages required for the unit tests).

This might not be the most elegant approach, and I'm unsure if it follows the intent of `docker/*`. Perhaps @gizmoguy and/or @rsanger could weigh in here? I'm happy to re-work this to make sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,9 @@
 *pyc
+
+dist
+
+ryu_faucet.egg-info
+
+docker/ryu-faucet-dev.tar.gz
+
+docker/tests
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,9 @@ PROJECT_NAME = ryu_faucet
 DIST_DIR = dist
 SRC_DIR = src
 
+## Used by the development Dockerfile
+DOCKER_DEV_PKG = ryu-faucet-dev.tar.gz
+
 all: clobber sdist uml
 
 uml:
@@ -31,6 +34,11 @@ sdist:
 	@echo Building Python package installable via "pip"
 	$(MKDIR) $(DIST_DIR)
 	$(PYTHON) setup.py sdist
+
+dockerdev: sdist
+	@echo Copying Python package to docker/${DOCKER_DEV_PKG} for use by development Dockerfiles
+	cp ${DIST_DIR}/ryu-faucet-*.tar.gz docker/${DOCKER_DEV_PKG}
+	cp -r tests docker/
 
 codefmt:
 	@echo Run below command manually to inline replace current code with newly formatted code per “pep8” guidelines
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,0 +1,21 @@
+FROM osrg/ryu
+
+RUN \
+  apt-get update && \
+  apt-get install -qy --no-install-recommends python-pip
+
+ADD ryu-faucet-dev.tar.gz /
+
+RUN \
+  pip install ipaddr; \
+  pip install /ryu-faucet-dev.tar.gz
+
+VOLUME ["/config/"]
+WORKDIR /usr/local/lib/python2.7/dist-packages/ryu_faucet/org/onfsdn/faucet/
+ENV FAUCET_CONFIG=/config/faucet.yaml
+ENV FAUCET_LOG=/config/faucet.log
+ENV FAUCET_EXCEPTION_LOG=/config/faucet-exception.log
+
+EXPOSE 6633
+
+CMD ["ryu-manager", "--verbose", "--ofp-tcp-listen-port=6633", "faucet"]
--- a/docker/Dockerfile.tests
+++ b/docker/Dockerfile.tests
@@ -1,0 +1,24 @@
+FROM osrg/ryu
+
+RUN \
+  apt-get update && \
+  apt-get install -qy --no-install-recommends python-pip vlan psmisc git sudo net-tools iputils-ping netcat-openbsd tcpdump
+
+RUN \
+  git clone git://github.com/mininet/mininet; \
+  cd mininet; \
+  git checkout -b 2.2.1 2.2.1; \
+  cd ..; \
+  mininet/util/install.sh -nfv
+
+COPY ryu-faucet-dev.tar.gz /
+COPY tests/ /tests/
+COPY runtests.sh /
+
+RUN \
+  pip install ipaddr; \
+  pip install /ryu-faucet-dev.tar.gz
+
+ENV FAUCET_DIR /usr/local/lib/python2.7/dist-packages/ryu_faucet/org/onfsdn/faucet/
+
+CMD ["/runtests.sh"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,6 +1,6 @@
 ## Faucet Dockerfile
 
-This directory contains two docker files **Dockerfile** and **Dockerfile.gauge**.
+This directory contains four docker files: **Dockerfile**, **Dockerfile.dev**, **Dockerfile.tests**, **Dockerfile.gauge**:
 
 ### Dockerfile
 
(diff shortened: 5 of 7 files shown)


gizmoguy:
Hey Simeon,

Thanks for the work. I think the way you've gone about this is fine.

One suggestion is that we drop the Dockerfile.dev and make that the main dockerfile. The reason I suggest this is that our Docker auto build system supports converting branches to "Docker tags" so we don't actually need to use the pip packages at all anymore.

If you're keen to, modify Dockerfile and Dockerfile.gauge to use "make dockerdev" instead of pip and then drop Dockerfile.dev (otherwise I can do it).

When this gets merged in I'll add a faucet/faucet-tests docker to our docker organisation (https://hub.docker.com/u/faucet/)

gizmoguy:
Actually another question, where does ryu-faucet-dev.tar.gz come from? I don't see any code to generate it in this PR

gizmoguy:
Actually just noticed @anarkiwi jumped the gun a little and merged this already. @simeonmiteff perhaps create a new PR with those changes I mentioned...

simeonmiteff:
Thanks for the feedback, @gizmoguy

To answer your question, `ryu-faucet-dev.tar.gz` is created by the the dockerdev `make` target (added by commit 84c09da27b256058bdf5653db1e3ae36f8143ba1). It builds the pip-installable package (by depending on sdist) and then copies `dist/ryu-faucet-*.tar.gz` to `docker/ryu-faucet-dev.tar.gz` (so that `Dockerfile.dev` can install package with a stable file name that exists inside the Docker build context).

Now, I don't understand how to get Docker tags to match faucet's git branches/tags instead of always downloading the current pip package version (as it does now).

My current (_and admittedly sketchy_) understanding is that doing so requires the build to pull (or maybe `wget`) the code for that specific branch/tag from GitHub and then `make`+install faucet. Besides being wasteful (because faucet already lives _next to_ `docker/`), there is the dilemma of referencing the tag/branch in the Dockerfile, which is itself versioned. The build can't `make dockerdev`, because that requires files out of the context.

Enabling automatic builds this way also conflicts with my objective to allow builds using local uncommitted code during development (what I was trying to achieve with the dockerdev `make` target and `Dockerfile.dev`).

How obnoxious is the idea of moving the Dockerfiles into the top level, so that the whole project is in the Docker build context? This would solve both of these problems.

gizmoguy:
Yeah, right, I missed where it depended on sdist. That looks good to me now.

The tag system doesn't really work at the moment because it will always pull the latest version from PIP, something I've been meaning to fix and then I saw this PR come up where you've done most of the work already.

I'm fine with the moving Dockerfile, Dockerfile.gauge, Dockerfile.test to the root of the project to make building easier. Also we should make all the Dockerfiles use the new "make dockerdev" target so that each individual build gets the checked out code rather than pip code. I think we should leave the config files in the docker directory there to keep the root of the project clean.

I think doing what I propose will allow all use cases to work, local building with "docker build -t reannz/faucet:latest" will get local changes, and the autobuild system when it checks out all tags from git will also get the right version.

Anyone else want to sign off on the above? @anarkiwi @rsanger @KitL

anarkiwi:
I defer to my colleagues (I can't spell docker).

rsanger:
This all sounds good

I would propose one minor change -- moving only Dockerfile.dev and Dockerfile.test to root (and rename it to Dockerfile) and then getting rid of the dockerdist tar.gz step, instead `ADD ./ /faucet-src/` in the docker file. Allowing for local changes to be built and tested.

I think keeping the other automatically built Dockerfiles in a separate folder makes sense and reduces clutter, as mentioned these should be updated to checkout and pull the appropriate branch rather than using pip, at which point the Readme should be added to direct people to the automated build and essentially no should need to manually run these.

simeonmiteff:
Ok I've got a [new branch here](/simeonmiteff/faucet/tree/feature-docker-reorganise) that integrates KitL@217c6d972178de7da77fe344859a4da192a9a556 (from @KitL's PR #90) that moves Dockerfile.test into the root, and uses `setup.py develop` instead of pip. This doesn't require `make dockerdev` to work, instead copying `./` to `/faucet-src/` during the build, as per @rsanger's suggestion. I'm happy that this works, feel free to test the test :-)

Now I plan to move Dockerfile and Dockerfile.gauge to the root and have them use `setup.py install` or similar, and then drop the dockerdev make target entirely, and update the documentation.

@gizmoguy, @rsanger: do you need anything else to make the Docker auto builds work as intended?

Sorry for the spam, I'm checking in first, so that my next PR is perfect.

KitL:
Would it be possible to mount the src directory as a volume rather than
copying it in the docker build. That way you wouldn't need to rebuild the
docker container for every change you want to test?

rsanger:
@simeonmiteff the new branch is looking good, I've have a couple of minor comments.
1. `pip install ipaddr` can be dropped this was due to this not being a dependency in setup.py when the docker was originally created
2. `python setup.py develop` I assume this pulls the required dependencies? (optionally `pip install /faucet-src` can be used if this makes more sense)

Ok so I have had a talk to @gizmoguy about what the automated docker builds, it requires a docker file to build the current version of the repo. I.e. what Dockerfile.dev is doing anyway, (so ignore my comment about keeping a separate copy for autobuilds this is not needed moving everything to root makes sense). Sorry for the confusion on this point :(
3. So in terms of updating the docker for autobuilds, I think your docker/Dockerfile.dev wants to be updated to install faucet like ./Dockerfile.test and moved to the root directory (renamed simply as Dockerfile). docker/Dockerfile is no longer needed and docker/Dockerfile.gauge can have the same 
treatment as docker/Dockerfile.dev.

@KitL mounting src as a volume, could work I guess you would then need a script to install this directory at start-up and then run. However building a docker is pretty fast (run one command) so I'm not sure this is a huge advantage. This would not be ideal for the autobuilds though unless a default fallback was included.

@gizmoguy @KitL @anarkiwi @simeonmiteff  Does this all make sense to everyone?

vladimir-mencl-eresearch:
Hi,

I'm listening in to your conversation: just one bit about Docker building process.

Quoting from https://docs.docker.com/engine/reference/builder/ ..

````
Traditionally, the Dockerfile is called Dockerfile and located in the root of the context.
You use the -f flag with docker build to point to a Dockerfile anywhere in your file system.

$ docker build -f /path/to/a/Dockerfile .
````

So, you can build from the root of the github project (without having to copy the files anywhere), even if the Dockerfile sits somewhere else.

As for the question on mounting the directory as a volume to get live changes: the builder is building from a build context sent to the daemon and AFAIK cannot use mounts - just copies the whole build context to the daemon.

However, you can still build from what was in the source tree at the time of build, and then *run* the container with a volume mount to get live changes done to the working directory visible in the running container.

Hope this helps.

Cheers,
Vlad

gizmoguy:
I think everything above sounds great and @simeonmiteff and @rsanger are on the right track.

Can I throw another spanner in the works as well and suggest we replace the "gauge+faucet" docker with a docker that is just gauge. We've been having an internal email discussion about how we should be separating gauge from faucet due to performance concerns. If we add the two separate Dockers (faucet and gauge) we can use compose file (https://docs.docker.com/compose/compose-file/) to easily join them together.

Also we should really get a development mailing list for faucet for these kinds of discussions so no one gets left out :)

Brad

shivarammysore:
Faucet-dev@openflowsdn.org is the mailing list to use

/Shivaram
::Sent from my mobile device::

gizmoguy:
Fantastic, thanks @shivarammysore.

shivarammysore:
the mailing is also web browsable -
https://groups.google.com/a/openflowsdn.org/forum/#!forum/faucet-dev

PR #90, opened 2016-05-26
Tidy up the docker files and documentation slightly

KitL:
Main change is removing ENV statements and instead relying on the default
    locations for relevant files. We have default values for the locations
    of config and logging files, it makes sense for docker to use these also.

    This also allows us to have the logging and config in separate directories
    while using Docker.

    Adds the -p argument to the docker README code sections as this is
    almost certainly going to be needed by anyone deploying this.

    Also makes the README fit 80 character terminals.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,11 +7,8 @@ RUN \
 RUN \
   pip install ryu-faucet ipaddr
 
-VOLUME ["/config/"]
+VOLUME ["/etc/opt/faucet/", "/var/log/faucet/"]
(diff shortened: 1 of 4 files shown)


simeonmiteff:
Hi @KitL. Faucet's default for logs and configs are `/var/log/ryu/faucet` and `/etc/ryu/faucet`, respectively. I note you dropped the `ryu` part, with `/etc/opt/faucet/` and `/var/log/faucet/` in the Dockerfiles. This is not a major problem, but it might be nice to have consistency. Should we change the defaults in faucet.py?

KitL:
No, we should change the docker files.

Sorry that was a subtle merge collision I missed.

PR #94, opened 2016-05-29
fix the missing routes bug.

trungdtbk:
upon receiving arp reply, we only added the first route to forwarding table,
and ignore other routes that share the same gw.


anarkiwi:
1. Thanks for finding and fixing this!
2. To prove that the bug is fixed and stays fixed, please add a test (or
modify the existing test), so that two IP routes to the same gateway is
handled.

trungdtbk:
I've updated the unittest for this issue.

PR #95, opened 2016-05-29
Re-organise Docker support

simeonmiteff:
This starts with KitL@217c6d972178de7da77fe344859a4da192a9a556 (from @KitL's PR #90) and then implements most of the feedback on #88 from @gizmoguy, @rsanger, @KitL, @vladimir-mencl-eresearch, and fixes a few other problems I noticed:

- Updates Dockerfiles: `docker build` now copies the code and runs a local `pip install` for faucet.
- As a result of the above, I dropped the dockerdev make target and nuked Dockerfile.dev.
- This keeps the Dockerfiles in `docker/`, meaning the build command must be run in the root and specify `-f docker/Dockerfile` (or similar).
- Implements a handful of fixes related to @KitL updating of the log and config paths.
- Adds and removes some build dependencies in the Dockerfiles.
- Updates the documentation and fixes some spelling errors.


rsanger:
Looks perfect to me, just tested out the images. Thanks for all your hard work on this.

gizmoguy:
@simeonmiteff Thanks for the hard work.

Unfortunately this commit breaks our autobuild system. I'll work on a fix for this and put it in a new PR soon(ish).

Brad

PR #98, opened 2016-05-31
Configure switch port number if faucet config of test cases

trentindav:
- created hw_switch_config.yaml to provide all the info required when using a HW switch
- dataplane ports can be specified and will be used when creating the FAUCET config file to use
  in each test. The PC interface name connected to these ports is not used at the moment, only the
  Faucet config string of each test is updated

Is this configuration approach acceptable?


KitL:
I am a bit concerned about the fact that the test configs are supposed to document the correct config for each feature. Using the string formatting in these configs makes it harder for non-python-fluent people to understand how to configure things.

I think we might have to have a think about another way of either making this work or another way of documenting configuration for features.

anarkiwi:
That's true; there is the expectation that people can at least read code.

I don't think this is a concern because (IMHO) SDN engineers will be
compatible with that expectation (and it is a pretty low bar).

Separately, it would be great to have an actual non-code based manual, but
given this is an open source project.... we must allocate scarce resources.

Just making a number of minor fixes here to restore tests (my fault, I
should have requested them from you before merging).

anarkiwi:
OK, all tests are now restored.

trentindav:
Sorry about the bad string substitution, I pushed the change too quickly

anarkiwi:
No problem. I am breaking out changes one by one you made in your test
script and adding them (for example, just added the OF1.3 flow-style
matches). Should be fully integrated today.

PR #110, opened 2016-06-09
Use Ryu ofctl_rest API to dump flows instead of ovs-ofctl in unit tests

trentindav:
Not all HW switches can support ovs-ofctl to dump all the flows installed in its tables.
By using the already connected Ryu controller to fetch the flows (by starting it not only with the faucet app but also with the ofctl_rest app) we can be able to remove this contraint.
The format of the strings to search in the "dump-flows" output has been modified to correspond to the output provided by the Ryu ofctl_rest application.

All tests have been successfully run on Mininet with this change applied.

### DIFF
--- a/tests/faucet_mininet_test.py
+++ b/tests/faucet_mininet_test.py
@@ -189,11 +189,6 @@ hardware: "%s"
         self.one_ipv6_ping(host, self.CONTROLLER_IPV6)
 
     def wait_until_matching_flow(self, exp_flow, timeout=5):
-        # TODO: actually verify flows were communicated to the physical switch.
-        # Could use size of ofchannel log, though this is not authoritative.
-        if SWITCH_MAP:
-            time.sleep(1)
-            return
         for _ in range(timeout):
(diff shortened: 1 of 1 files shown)


tests/faucet_mininet_test.py
```diff
@@ -185,22 +188,21 @@ def one_ipv6_ping(self, host, dst):
     def one_ipv6_controller_ping(self, host):
         self.one_ipv6_ping(host, self.CONTROLLER_IPV6)
 
-    def wait_until_matching_flow(self, flow, timeout=5):
+    def wait_until_matching_flow(self, exp_flow, timeout=5):
         # TODO: actually verify flows were communicated to the physical switch.
```
anarkiwi: Please remove this TODO and the if SWITCH_MAP: here, because your very nice change fixes the problem for hardware also.

anarkiwi:
This is a great change!

I added one note to remove the TODO/sleep for hardware switches because
your change fixes the problem.

anarkiwi:
I went ahead and merged this and I will make some further changes myself to ensure works for hardware too.

PR #119, opened 2016-07-20
Auto code fixing with QuantifiedCode	

shivarammysore:
Fixed code to use best practice per QuantifiedCode


anarkiwi:
Just so I as I make sure. I presume the unit tests pass, with this change?

PR #129, opened 2016-07-27
get fib installed immediately when the nexthop is already resolved

trungdtbk:


anarkiwi:
This is a nice change.

Few things:

+        for controller_ip in vlan.controller_ips:
+            if ip_gw in controller_ip:
+                is_gw_valid = True
+                break
+            else:
+                continue

else:
   continue

is not necessary.

When you add/del a route, you check to see if any flows should be sent -
you can move the retrieval of the ryu DP at the same time.
flowmods = [] needs to move out a block so it is always defined.

Does it pass all tests? I think flowmod not being defined always would be a
problem.

trungdtbk:
My intention is to check if the if the route is valid before adding the route to the route tables.

anarkiwi:
Totally get the intention. I like it!

I'm talking about how it's done. :)

you can see, that flowmods = [] is defined in an inner block. It should be
moved out a block, so it's always defined.

trungdtbk:
I fixed that. And it passes all the tests.

anarkiwi:
Excellent, I merged your change. Thank you for your contribution.

PR #133, opened 2016-07-29
Fix a (serious) mistake: missing fib_table argument in add_route

trungdtbk:
This shouldn't have passed the test :)
I'm trying to figure out how to write a unitest for add_route (when next-hop is resolved) and del_route
I added "process" to exabgp to program the bgp updates, but it does not work. Somehow, exabgp says "permission denied" when the control script is placed inside a temp dir.

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/valve.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/valve.py
@@ -786,16 +786,18 @@ class Valve(object):
             routes = vlan.ipv6_routes
             neighbor_cache = vlan.nd_cache
             eth_type = ether.ETH_TYPE_IPV6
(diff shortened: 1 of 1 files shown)


anarkiwi:
That's right; it should not have. I see lint errors there too. 
I will investigate why the test passes when it should not as that must mean the routing tests are broken. Thanks for picking this up and submitting a fix quickly.

trungdtbk:
It passed because exabgp sent route announcements too quick so that none of them has nexthop resolved. If we can delay the announcement of 10.0.3.0/24, we could see it fail.

anarkiwi:
I noticed this also. We should also check that the switch has the correct
flows installed with the resolved MAC addresses, which is what I am adding
now.

PR #147, opened 2016-08-01
Updated feature list

shivarammysore:


anarkiwi:
" Policy based forwarding
￼ +- 802.1x authentication support"

Suggest to say instead:

 policy based forwarding to offload processing to external systems (Eg 802.1x via hostapd).

Makes it clearer that 802.1x is not and does not need to be in the controller.

shivarammysore:
Good suggestion.  Fixed and updated Pull request.

Thanks!

PR #149, opened 2016-08-04
Config consistency

KitL:
Here is the config update.

I havent changed most of the documentation, so these are not really the official config. However, we should move to deprecating the old format quickly.

See commit messages for further details.

### DIFF
--- a/Dockerfile.gauge
+++ b/Dockerfile.gauge
@@ -2,51 +2,17 @@ FROM osrg/ryu
 
 RUN \
   apt-get update && \
-  apt-get install -qy --no-install-recommends wget python-pip supervisor \
-    apt-transport-https libyaml-dev libpython2.7-dev
+  apt-get install -qy --no-install-recommends python-pip \
+    libyaml-dev libpython2.7-dev
 
 COPY ./ /faucet-src/
 
 RUN \
-  pip install wheel && \
-  pip install influxdb && \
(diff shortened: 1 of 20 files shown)


anarkiwi:
I patched this into my own repo (is current with yours), and looks like
some typos?

root@agent-p:~/faucet/tests# ./faucet_mininet_test.py

required binary/library ryu-manager --version version is 4.4

required binary/library ovs-vsctl --version version is 2.3

required binary/library tcpdump -h version is 4.5

required binary/library nc present (nc from the netcat-openbsd)

required binary/library vconfig present (the VLAN you are talking about)

required binary/library fuser -V version is 22.2

required binary/library mn --version version is 2.2

required binary/library exabgp --version version is 3.4

required binary/library pip show influxdb version is 3.0

required binary/library pylint --version version is 1.6

No config file found, using default configuration

No config file found, using default configuration

************* Module org.onfsdn.faucet.gauge

E:141,38: Undefined variable 'sef' (undefined-variable)

E:147,37: Undefined variable 'ryudp' (undefined-variable)

E:148,26: Undefined variable 'ryudp' (undefined-variable)

lint of gauge.py returns an error

pylint must pass with no errors

anarkiwi:
Can I suggest also - you have some nicely separable changes (Eg the docker change). Perhaps they could be submitted as part of a separate PR?

PR #150, opened 2016-08-07
Config fix

KitL:
Sorrhy I havent split that. If you have an issue with this PR then I'll break it up, but I dont think the other parts are valuable enough on there own to split them into separate branches.

### DIFF
--- a/Dockerfile.gauge
+++ b/Dockerfile.gauge
@@ -2,51 +2,17 @@ FROM osrg/ryu
 
 RUN \
   apt-get update && \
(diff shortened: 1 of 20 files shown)


anarkiwi:
I don't have any issue with the PR; it's a matter of risk management.

I just make the suggestion, that you have a large series of small PRs,
rather than a small series of large  PRs, it's much easier to assess and
manage risk, especially when a small PR just fixes (for example) a docker
definition. And a small (in terms of lines of code) fix like what you made
there, is very valuable.

I think it will make it easier, especially for you is all :-).

PR #162, opened 2016-08-10
Add multiple DP support to Faucet

Callum027:
Here it is. The gist of it is that it creates a mapping between DPID and Valve objects (which are still separate for each DP), and modifies every function which handles DPs in order to take this into account, and handle appropriately (e.g. run for every DP in `reload_config`, get the DPID and work with that DP's valve object in `_packet_in_handler`).

There is also a new test base class for testing multiple DPs. At the moment, there are very basic implementing tests (a `pingall`), but these can easily be changed if they're not adequate.


anarkiwi:
Very good. I'll try it right now. My home production deployment currently
requires two FAUCETs so I will know very quickly if this works!

anarkiwi:
Good news. I was able to move my Zodiac to use the same FAUCET as my AT and
I merged the configs.

Works a treat.

I'll continue testing before I add feedback.

Good stuff,

PR #185, opened 2016-08-30
Database support for Gauge

shivarammysore:


shivarammysore:
Rahul tested the entire setup to work correctly.  @shivarammysore reviewed the architecture.

shivarammysore:
This code was developed as  a part of the SDNhackfest project (www.SDNHackfest.org).

KitL:
Structurally I am concerned about how the code is being added to Gauge. Gauge should just throw updates at watchers, and watchers are where the code for handling pushing to the database should be. IE gauge shouldnt need to import anything to do with the database.

Also, you should build new watchers for use with couch db rather than modifying existing ones to also send stats to couch. At the moment this will crash in the REANNZ office because we are not using couchdb so when it tries to open its config file I'll get an Exception. You need to leave the existing watchers as they are.

Furthermore the config should come in through config.py. watcher.py shouldnt be parsing config. I would also suggest that you should just use the new config format, so your db config can all be included in gauge.yaml.

So structurally the only files you should really need to touch is watcher.py and config.py. There may also be some minor changes to how gauge.py handles adding and removing dps so that your watchers can monitor that effectively. But mostly I would expect to see that you would add a GaugeFlowTableCouchPoller class and maybe a GaugeDPCouchLogger.

I'll add inline comments to sorta clarify what I mean.

src/ryu_faucet/org/onfsdn/faucet/gauge.py
```diff
@@ -96,6 +101,36 @@ def __init__(self, *args, **kwargs):
...
+                    init_switch_db(self.switch_database)
+
+                self.flow_database, exists = self.conn.create(
+                                                data['flows_doc'])
+                # Create database specific views for querying
+                if not exists:
+                    init_flow_db(self.flow_database)
+                self.db_conf_data = data
+
+
```
KitL: Reading the config should be done in config_parser.py, then the parsed config should be giving to a watcher, and any initialising of the database can be done when that watcher is initialised. If you need to have that happen only once, then let the watcher call some kind of function in nsodbc.py that ensures that.

src/ryu_faucet/org/onfsdn/faucet/gauge.py
```diff
@@ -108,6 +143,17 @@ def handler_connect_or_disconnect(self, ev):
...
+            if self.db_enabled:
+                rows = self.switch_database.get_docs(
+                    self.db_conf_data['views']['v1'],
+                    key=str(hex(ryudp.id))
+                )
+                if not rows:
+                    switch_object = {'_id': str(hex(ryudp.id)),
+                                     'data':{'flows':[]}}
+                    self.switch_database.insert_update_doc(switch_object,
+                                                           'data')
```
KitL: this should be a watcher's job. This might take some modification to get this actually to work tidily, but I kinda suspect it should be fairly straightforward.

src/ryu_faucet/org/onfsdn/faucet/gauge.py
```diff
@@ -116,6 +162,19 @@ def handler_connect_or_disconnect(self, ev):
...
+                rows = self.switch_database.get_docs(
+                    self.db_conf_data['views']['v1'],
+                    key=str(hex(ryudp.id))
+                )
+                switch = rows[0].value
+                # Delete flows in the switch
+                for flow_id in switch['data']['flows']:
+                    self.flow_database.delete_doc(str(flow_id))
+                # Delete switch from database
+                self.switch_database.delete_doc(str(hex(ryudp.id)))
```
KitL: same as above.

src/ryu_faucet/org/onfsdn/faucet/watcher.py
```diff
@@ -35,6 +37,38 @@ def watcher_factory(conf):
...
+    Initialize/refresh switch database
+    Args:
+        switch_database
+    """
+    views = {}
+    views["switch"] = {}
+    views["switch"]["map"] = "function(doc) " + \
+                             "{\n  emit(doc._id, doc);\n}"
+    switch_database.create_view("switches", views)
+
```
KitL: Could these functions exist  in nsodbc.py? I think that would be tidier.

src/ryu_faucet/org/onfsdn/faucet/watcher.py
```diff
@@ -301,6 +335,38 @@ class GaugeFlowTablePoller(GaugePoller):
...
+                    init_switch_db(self.switch_database)
+
+                self.flow_database, exists = self.conn.create(
+                    data['flows_doc'])
+                # Create database specific views for querying
+                if not exists:
+                    init_flow_db(self.flow_database)
+                self.db_conf_data = data
+        # TODO: DB update counter can be made configurable
+        self.db_update_counter = 5
```
KitL: This should be in a GaugeFlowTableCouchPoller class, rather than this one. Also I think when this is initialised is a more appropriate time to initialise the database.

KitL:
I hope that mostly makes sense.

shivarammysore:
It would be better if the architecture and model is documented in
architecture.rst file so that it is clear.

Thanks

PR #265, opened 2016-10-21
Updated class diagram and added version info for gauge.yaml file

shivarammysore:


shivarammysore:
Thanks to @rahul217 for testing Gauge with the new changes.

PR #273, opened 2016-10-27
Setting up Open vSwitch on QOTOM and making it work with Faucet

shivarammysore:


anarkiwi:
Nice!! Thanks for this!!

PR #275, opened 2016-11-01
Incrementally reload faucet config

trungdtbk:
incrementally reloading Faucet configs
Detects what changes have been made and reload only what affected by the changes.
Following changes are detectable: 
- port configs change such as native_vlan, unicast_flood, acl_in....
- ACL changes
- VLAN changes: controller_ips....

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/dp.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/dp.py
@@ -16,6 +16,7 @@
 from conf import Conf
 from vlan import VLAN
 from port import Port
+from valve_acl import ACL
 
 import networkx
 
@@ -132,7 +133,7 @@ class DP(Conf):
 
     def add_acl(self, acl_ident, acl_conf=None):
         if acl_conf is not None:
-            self.acls[acl_ident] = [x['rule'] for x in acl_conf]
+            self.acls[acl_ident] = ACL(acl_ident, acl_conf)
 
     def add_port(self, port):
         port_num = port.number
@@ -271,7 +272,7 @@ class DP(Conf):
 
         def resolve_port_names_in_acls():
             for acl in self.acls.itervalues():
-                for rule_conf in acl:
+                for rule_conf in acl.rules:
                     for attrib, attrib_value in rule_conf.iteritems():
                         if attrib == 'actions':
                             if 'mirror' in attrib_value:
--- a/src/ryu_faucet/org/onfsdn/faucet/port.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/port.py
@@ -50,7 +50,7 @@ class Port(Conf):
         self._id = _id
         self.update(conf)
         self.set_defaults()
-        self.phys_up = False
+        self.dyn_phys_up = False
 
     def set_defaults(self):
         for key, value in self.defaults.iteritems():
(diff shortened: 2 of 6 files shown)


anarkiwi:
Thanks for this!

1. We need tests for this. An example of one of several tests - supply one
config with an ACL, then update the config, and check that the config has
changed.

2. Use something like hash(frozenset(self.__dict__.iteritems())) to hash on
instance variables, so that the list is not hardcoded.

trungdtbk:
The PR already has a simple test. It changes native_vlan of a port and HUPs Faucet, then checks if config change is made and whether durations of other flow increases.

I tried using self.__dict__, but there is an issue. Some of instance variables are dynamic, e.g. ipv4_routes,  arp_cache and phys_up, which cannot be included in the hash (and not part of configs). Otherwise, config changes always result in true.

anarkiwi:
They PR does not contain any tests:


   - *M* src/ryu_faucet/org/onfsdn/faucet/dp.py
   <https://github.com/REANNZ/faucet/pull/275/files#diff-0> (5)
   - *M* src/ryu_faucet/org/onfsdn/faucet/port.py
   <https://github.com/REANNZ/faucet/pull/275/files#diff-1> (12)
   - *M* src/ryu_faucet/org/onfsdn/faucet/valve.py
   <https://github.com/REANNZ/faucet/pull/275/files#diff-2> (93)
   - *M* src/ryu_faucet/org/onfsdn/faucet/vlan.py
   <https://github.com/REANNZ/faucet/pull/275/files#diff-3> (19)

Perhaps a missed check in? If so please update the PR.

It would be good to rename the the dynamic attributes in a way you can
easily filter them out (eg with a prefix). Ideally they should be a
subclass but I don't want to hold up your work to do that. It's very
important not to hardcode the list in the hash code calculation.

trungdtbk:
My mistake. I've updated the RP with a commit containing the test.
I understand. I will look at that later.

anarkiwi:
Just following this up; have you had a chance to take another look at attribute hashing?
It would be great to have your PR integrated!

trungdtbk:
Hi Josh,
I have revised the PR, and added unittest for the changes.

anarkiwi:
This is a great step forward. Well done.

For detecting differences, it might be better to use python's set
operations (since they can compute difference/unions etc so you don't have
to build lists and diff yourself). It would also be nice to explicitly
tests that if you change an ACL (for example) that the rules are present
(you do test behavior which is great, but explicitly verifying rules
changing would be good too).

Nonetheless, this is a great incremental step forward. Thank you very much
for your contribution to the project.

trungdtbk:
I don't really get it how to use set operation now. I will study it.
The current tests can verify if a rule is added, it will have to exist in the sw. I may add a test for adding/removing acl_in in a port.
In future, each rule in ACL should have its own priority and id, so that we don't need to use the order they appear. And, we can add a new rule without reloading others. I will look at this issue.

anarkiwi:
Those would be great next steps.

Thanks!

anarkiwi:
I see the config test is failing:

root@agent-p:~/faucet/tests# ./test_config.py

E..............

======================================================================

ERROR: test_acl (__main__.DistConfigTestCase)

  File "./test_config.py", line 217, in test_acl

    self.assertEquals(dp.acls[dp.ports[1].acl_in][0]['nw_dst'], '172.0.0.0/8
')

TypeError: 'ACL' object does not support indexing


Ran 15 tests in 0.760s

trungdtbk:
I will check that

PR #290, opened 2016-11-13
Routing

trungdtbk:
Use GROUP table for FIB and VLAN flooding.

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/dp.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/dp.py
@@ -45,6 +45,7 @@ class DP(Conf):
     high_priority = None
     stack = None
     ignore_learn_ins = None
+    group_table = False
 
     # Values that are set to None will be set using set_defaults
     # they are included here for testing and informational purposes
@@ -89,6 +90,7 @@ class DP(Conf):
         # Flooding will still be done by the dataplane even with a packet
         # is ignored for learning purposes.
         'ignore_learn_ins': 3,
+        'group_table': False,
         }
 
     def __init__(self, _id, conf):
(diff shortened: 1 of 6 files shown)


anarkiwi:
1. This is another great move forward.

2. Have you tested it on hardware? I am pretty certain this will not work
on Zodiac. I tried before on AT with an earlier patch and it did not work.

3. Either way, it would be good to have use of the group table be a
configurable option.

trungdtbk:
I haven't tested on hardware. I will try today. I just think group ALL is an OF required feature so it should work on hardware.
We can't believe OF compliance certificate  (?)

I will turn it into a configurable option.

anarkiwi:
You are right, the compliance certificate is worthless. I only trust
running code on a real network.

Thanks,

trungdtbk:
I tested on Zodiac and it doesn't work. The worse part is that Zodiac goes silent.
If a switch cannot commit a flow, it should give the controller an error. With Zodiac, everything seems completely OK. Flows do exist but it doesn't create the requested groups and drops traffic. I'm going to test on Allied Telesis as well.

anarkiwi:
Thanks for the update! Yes, not the best error handling. But sadly confirms
suspicion it would not work on Zodiac.

trungdtbk:
I've updated the PR. GROUP table is used for both flooding and FIB.
Group table is now optional and can be turned on by adding "group_table: True" to a dp config. The default is off.
It's quite complex to use GROUP table for implementing stacking logic, so GROUP table will be disabled if stacking is on.

anarkiwi:
Ping - has this been tested on hardware?

trungdtbk:
I tested it working on AT-510X, but not Zodiac.

anarkiwi:
Great - please would you update the PR and let us know if it passes tests
under docker?

If it does keen to get it in.

trungdtbk:
I will update it and get back to you soon.

PR #294, opened 2016-11-18
Optimization to install scripts

shivarammysore:


gizmoguy:
Looking far more sensible and usable now.

Thanks for making the requested changes.

PR #333, opened 2016-12-22
Vlan acls

KitL:
I am not sure why there are changes to wait_debug_log it seems like it is changing it to what it already is.

But this has working tests (aside from lldp) for adding vlan acls.

new table structure goes: port_acl, vlan, vlan_acl, eth_src, etc.


anarkiwi:
I'll take a look; does it work on hardware (specifically AT? Will test it
on the ATs I have but would good to know for sanity checking)

anarkiwi:
I tested this on hardware and confirmed I could still get full port
utilization. Great!

PR #350, opened 2017-01-06
Fix testcase for LLDP. Add testcase for CDP.

agarwalpiyush:
Ran docker tests.


anarkiwi:
You will want to check that the tests pass under docker (
https://github.com/REANNZ/faucet/blob/master/src/docs/README.docker.md).

ladvd when run under docker is missing a directory that it wants; see the
LLDP test for what it does.

PR #351, opened 2017-01-07
convenient scripts to setup Faucet service on Ubuntu

shivarammysore:


shivarammysore:
based on @anarkiwi blog post https://faucet-sdn.blogspot.com/2016/10/running-faucet-as-systemd-controlled.html

PR #363, opened 2017-01-12
Add a README for using Faucet on Aruba switches

abhayb-hpe:
I have written a small readme on how to go about using Faucet with Aruba switches. It talks about how to configure both the Aruba switch and faucet to make it work.

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/aruba/README.md
+++ b/src/ryu_faucet/org/onfsdn/faucet/aruba/README.md
@@ -1,0 +1,206 @@
+# FAUCET on HPE-Aruba Switches
+
+
+## Introduction
+All the Aruba's v3 generation of wired switches support the faucet pipeline.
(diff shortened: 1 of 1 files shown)


shivarammysore:
@abhayb-hpe can you move this to src/docs/ directory and name this file as Readme_Aruba.rst file?  This makes our directory structure clean

abhayb-hpe:
@shivarammysore - I can do that. Is it ok if I move it to src/docs and name it as README.Aruba.md? I see similar naming conventions for the other README files in the directory.

shivarammysore:
You can certainly do that, but with a name - README_Aruba.md or
 README_Aruba.rst (preferred).  At some point in time, we need to move from
md to restructuredText (rst) format as it be easy to build documentation.

Thanks

abhayb-hpe:
README_Aruba.rst it is then.

PR #367, opened 2017-01-12
Move the readme file of Aruba to the docs directory

abhayb-hpe:
Moving the file to src/docs/ based on the suggestion from @shivarammysore .


shivarammysore:
Thanks.  It will look lot more nicer if you could fix some markup.  Here
are some references:

   1. http://docutils.sourceforge.net/docs/user/rst/quickref.html
   2.
   https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst
   3. https://wiki.typo3.org/Editors_(reST)
   4. Sphinx can also be used to test rst to html -
   http://www.sphinx-doc.org/en/1.5.1/tutorial.html - you will need these
   kinds of tools as you will not be able to see issues till it gets published.

abhayb-hpe:
Thanks for letting me know. I didn't realize md and rst have totally different styles.
I have made the required changes to use the rst markup language in the document.

PR #376, opened 2017-01-17
use GROUP tables for FIB and VLAN flooding

trungdtbk:
To enable, add "group_table: True" to each dp's config.
group table is automatically disabled if stacking is used.


anarkiwi:
Thanks so much for your fast work updating this!

At the moment are coming up to NZNOG, so are focussed on fixing any show stopper bugs that might affect the deployment. We will look at this change in more detail after that. I'm sorry for the delay.

PR #377, opened 2017-01-18
Correct table numbering.

anarkiwi:


shivarammysore:
Would it be a good idea to add Group Table info to this?

anarkiwi:
Certainly, once merged and tested (we are in a freeze heading into NZNOG).

PR #380, opened 2017-01-24
use TERM instead of KILL

isomer:
ryu tends to squash errors, and thus not exit.  This wrapper
KILL's itself when it gets exceptions in various places.

Change it so it uses TERM instead of KILL thus messages are cleaned up
cleanly


gizmoguy:
Please set git author.name and author.email.

gizmoguy:
Please remove extraneous commits from this PR.

anarkiwi:
I merged just the named changes in a new commit.

https://github.com/REANNZ/faucet/pull/384/commits/41a6e94ad59ca73ba1bf982a6336adbdbecbc08b

PR #382, opened 2017-01-26
Ratelimit relearning the same L2 host to 1/s.

gizmoguy:


gizmoguy:
Do we want to make this configurable on a DP?

gizmoguy:
Apparently @isomer and @KitL have a better solution for this.

PR #393, opened 2017-02-05
avoid adding invalid BGP routes when vlan having multiple controller ips

trungdtbk:


anarkiwi:
Thanks for checking this!

We need a test to make sure a bad route doesn't get in.

It would be good not to use an anonymous variable "flag". Can you add another function and move the validating code there?

trungdtbk:
I've updated this RP. Unittest added.

anarkiwi:
Very nice! Thank you.

PR #475, opened 2017-03-21
add barrier after group_mod add/delete

trentindav:
When Groups are used the controller needs to send a barrier_request to make sure that:
- the group is deleted before adding other groups (in case the same group_id is deleted and added again)
- the group is installed in the switch before the following flow_mods (which are using this group) are processed.

I added a barrier after each `ofmsgs.append(valve_of.groupdel/groupadd)` I found in valve.py, valve_flood.py, and valve_route.py

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/aruba/aruba_pipeline.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/aruba/aruba_pipeline.py
@@ -1,7 +1,6 @@
 import inspect
 import json
 import os
-import sys
 
 from ryu.ofproto import ofproto_v1_3
 from ryu.ofproto import ofproto_v1_3_parser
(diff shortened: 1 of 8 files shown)


anarkiwi:
Thanks!

Can I suggest instead to use the same strategy in valve_flowreorder()? ie.
don't explicitly add barriers, but move the groupdels to the head and then
push one barrier?

trentindav:
Sure, it makes more sense, I didn't see that function before.
I'll remove these barriers and add them only in valve_flowremove(), so that this function reorders the list of message to finally have:
1. flowdels and groupdels
2. barrier
3. groupadd
4. barrier
5. everything else

trentindav:
I modified the code as described above, but now I see a problem when running test cases with groups:
```
$ sudo python faucet_mininet_test.py FaucetGroupTableTest
....
======================================================================
ERROR: __main__.FaucetGroupTableTest.test_group_exist (subunit.RemotedTestCase)
__main__.FaucetGroupTableTest.test_group_exist
_StringException: Traceback (most recent call last):
  File "/home/davide/faucet/tests/faucet_mininet_test_base.py", line 257, in tearDown
    msg='debug log has OFPErrorMsgs')
  File "/usr/lib/python2.7/unittest/case.py", line 416, in assertFalse
    raise self.failureException(msg)
AssertionError: debug log has OFPErrorMsgs
```

I see that OVS is generating OFPET_GROUP_MOD_FAILED/OFPGMF_GROUP_EXISTS errors because the controller is deleting and adding the same group more than once:
- before flowreorder: groupdel, groupadd, groupdel, groupadd (always same group_id, the test does it for group_id=100 and 4196)
- after flowreorder: groupdel, groupdel, groupadd, groupadd (this last add causes the error because the group is already there)
- with my first change (barrier after each groupdel/groupadd): groupdel, barrier, groupadd, barrier, groupdel, barrier, groupadd, barrier

Do you know why the same group is added/deleted more than once?
Do you think that we should check for duplicate group_ids in valve_flowreorder?

anarkiwi:
It will send the add/delete pair more than once due to less than optimal
provisioning.

For now yes dropping a duplicate group add is OK (and add a TODO saying we
should optimize further back). There will be other situations where can see
we likely do this.

PR #485, opened 2017-03-26
use flowremoved to re-learn hosts instead of hard_timeout

trungdtbk:
I propose to use OFFlowRemoved to expire and re-learn hosts instead of hard_timeout.
I experienced performance loss when on AT switches due to something wrong with the combination of (OUTPUT_CONTROLLER and GOTO_TABLE). Faucet can capture flowremoved events and removes expired rules.

This patch is for release V1_4_0

### DIFF
--- a/src/ryu_faucet/org/onfsdn/faucet/faucet.py
+++ b/src/ryu_faucet/org/onfsdn/faucet/faucet.py
@@ -560,6 +560,44 @@ class Faucet(app_manager.RyuApp):
 
         self._send_flow_msgs(ryu_dp, flowmods)
 
+    @set_ev_cls(ofp_event.EventOFPFlowRemoved, MAIN_DISPATCHER) #pylint: disable=no-member
+    @kill_on_exception(exc_logname)
+    def handler_flowremoved(self, ryu_event):
+        ofmsgs = []
+        msg = ryu_event.msg
+        ryu_dp = msg.datapath
+        dp_id = ryu_dp.id
+        ofp = msg.datapath.ofproto
+        valve = self.valves[dp_id]
+        if msg.reason == ofp.OFPRR_IDLE_TIMEOUT:
+            reason = 'IDLE TIMEOUT'
+            table_id = msg.table_id
+            match_oxm_fields = msg.match.to_jsondict()['OFPMatch']['oxm_fields']
+            eth_dst = None
+            vlan_vid = None
+            for field in match_oxm_fields:
+                if isinstance(field, dict):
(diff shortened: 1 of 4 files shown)


gizmoguy:
Trung, thanks for the patch.

Unfortunately you'll need to rebase this as I've just finished moving source files around in the repo.

Git should be able to do this automatically for you:

```
git remove add upstream https://github.com/REANNZ/faucet.git
git fetch upstream
git rebase upstream/master
git push -f origin v1_4_0
```

Second, I'm interested in knowing what performance limitations you're seeing? As perhaps the best course of action is for us to contact the vendor with a reproduction case instead of making modifications to our code. Though I'm interested in hearing other developers' thoughts on the merits of hard_timeout vs idle_timeout.

KitL:
Ok there is a large comment explaining the use of hard timeout and idle timeout that you will need to update as well: https://github.com/trungdtbk/faucet/blob/8647e1a2d2b6a581c257b967658f5fd2b12102ea/src/ryu_faucet/org/onfsdn/faucet/valve_host.py#L129

I think you have the behaviour backwards as well. The eth_src rule should have an idle timeout, and when that is removed you need to actively delete the eth_dst rule. So when we havent seen a packet from a host for a while we assume we no longer know where it is. Rather than when we havent seen someone try to talk to a host for a while we assume we no longer know where it is, even if it has been sending packets all the while.

This is also safer in the event that the delete gets lost (by ryu or by the switch). The switch sending packets to the incorrect location is better than the switch refusing to relearn the location but flooding packets regardless.

Also because I am here:
https://github.com/REANNZ/faucet/pull/485/files#diff-b1e21bb1b94a9158d41c17eb26c22a66R583
That should be an elif and the next line should use | instead of ^

trungdtbk:
I experienced packet loss, jitter and drop in TCP throughput. The main reason is:
when the hard_timeout expires and related rules in src_table get flushed, packets are output to the controller and also the dst_table. This combination of actions causes the issue.
I haven't tested on other hardware.

If we assume hardware is OK, there's still burden on the controller. The longer the latency of installation of fowmods, the bigger the burden is. If all hosts expire nearly at the same time, there will more packetins.

gizmoguy:
Ok gotcha, if you have lots of rules hard timeout at the same time then we have lots of load issues.

Sounds good to me, if you rebase and make @KitL suggested changes I will be more than happy to merge.

anarkiwi:
We may want to make this optional behavior as we don't know if it works on
all our vendors.

PR #486, opened 2017-03-27
add jitter to rule timeout to avoid re-learning a bunch of host simultaneously.

trungdtbk:
To enable hard_timeout, add 'hard_timeout_enabled: True' to dp config.
This PR is to replace PR #485

### DIFF
--- a/faucet/valve_host.py
+++ b/faucet/valve_host.py
@@ -16,9 +16,11 @@
 # limitations under the License.
 
 import time
+import random
 
 import valve_of
 
+LEARN_JITTER = 10 # seconds
 
 class HostCacheEntry(object):
 
@@ -125,7 +127,8 @@ class ValveHostManager(object):
                     self.eth_src_table, vlan=vlan, eth_src=eth_src),
                 priority=(self.host_priority - 2)))
         else:
-            learn_timeout = self.learn_timeout
+            #Add a jitter to avoid whole bunch of hosts timeout simultaneously
+            learn_timeout = self.learn_timeout + random.randint(0,LEARN_JITTER)
             ofmsgs.extend(self.delete_host_from_vlan(eth_src, vlan))
 
         # Update datapath to no longer send packets from this mac to controller


KitL:
You forgot to switch the eth_src and eth_dst tables. Or was that intentional?

trungdtbk:
Yes, I keep the use of idle_timeout is on dst_table. Users can disable/enable the hard_timeout on src_table.

KitL:
Unless you have a really strong reason for wanting it to be that way I would definitely prefer it for eth_src rules to have an idle timeout and eth_dst rules to be deleted only in response to eth_src timing out.

There's two reasons here:
- An eth_src rule timing out means the switch probably doesnt have a good idea where a device is (and therefore the rules should be deleted). An eth_dst rule timing out doesnt directly tell you anything about the location of the device in question, it's fairly likely that you dont know where the device is, but not necessarily.
- It is ok to have a rule in the eth_dst table without a corresponding rule in the eth_src table, because the switch will quickly relearn the mac address anyway. It is much worse to have the opposite situation, because the switch cannot learn the mac address again while that rule is there (and it wont get deleted). Sometimes openflow messages can get lost (particularly when ryu is trying to send too many messages at once). So it is safer to have the eth_src rule timeout and then delete the eth_dst rule than the other way around.

trungdtbk:
You're correct. It's really worse when openflow messages get lost. 
So in this case, we no longer need this to be optional. Faucet can relearn anyway even without hardware support for flowremoved acknowledgement. I will fix it now.

trungdtbk:
It's been updated

KitL:
We still need some way to time out rules in the eth_dst table when we arent using this option though. Now if the switch doesnt support flow removed messages, they will only get deleted when the mac address appears somewhere else in the network. So if a host is removed from the network its mac address will remain forever. I think we still need a configuration option for this.

Sorry to be such a nuisance!

trungdtbk:
I get your point. But I forgot why do we use hard_timeout?
If we use idle_timeout for both src and dst table, we can address the problem without the need for flowremoved msgs. By setting a higher value for the dst table, we can assure the src rules get removed first so the controller can re-learn and update both tables. If a host moves to a new port, it will be learned. The old rule will disappear automatically without the controller intervention.
We still can make the flowremoved msg optional so the controller can expire and remove inactive hosts from its memory. How do you think?

KitL:
Because with an idle timeout we cant guarantee that the eth_src rule will be deleted before the eth_dst rule. So you might have a device that sends just enough packets to keep the eth_src rule refreshed but the reply comes late enough that the eth_dst rule times out. Then all packets to that device will be flooded but the eth_src rule might never disappear.

This is probably the weirdest bit of faucet, hence the big comment.

trungdtbk:
It's getting more complicated now. This is the proposed behaviour that I would implement:
1) If the src rule exists, the corresponding dst rule exists
2) If the src rule expires, a flowremoved msg must be sent to controller. Then, controller compares in_port to see if the expired rule belong to a known host which has moved to another port. If no, it remove the dst rule. If yes, do nothing.
3) in case flowremoved msgs get lost, a host expire mechanism on the controller will query the src table, if no src rule for a host exists, a flowmod is sent to remove the rule in the dst table.
4) this is optional and enabled by default, so users can turn off if a switch does not support.
5) by doing this, we can handle all situations. 

If you're ok with that. I will proceed the implementation.

KitL:
So 
1. Right.
2. So the concern here is that the packet_in message could be received before the rule expiry message? Then you might delete the new output message created by the packet in? I guess we should just send the delete for the eth_dst table as well as a delete for the eth_src table. It would be good to be able to do this atomically, but I think we would be safe enough with this arrangement. I would suggest that giving eth_src rules a fairly large hard timeout might be a good idea (like 10x the idle timeout? + some random value to mitigate everything timing out at once).
3. I dont think this should be necessary. eth_src rules should always time out, and eth_dst rules should only be allowed to timeout when it is guaranteed the eth_src rule is not there. 
4. I am happy with that, but when this is disabled then the eth_src rule has a hard timeout and the eth_dst rule has an idle timeout of the same length, so we can guarantee the eth_src rule will be deleted first. it probably makes sense to give eth_dst a fairly large hard timeout as well.
5. In this case I am happy with things.

Also architecturally Valve should be the thing making the decision as to whether something should be deleted. Faucet should really only be extracting the relevant information to give to valve. I think Valve should have a rcv_rule_timeout(table_id, match) function.
Faucet determines it's a timeout and tells Valve. And then Valve decides whether this is a mac timing out or something else.

anarkiwi:
Looks like things have paused here.

Can I suggest a couple of ways forward (in addition to Chris' feedback especially on architectural questions - the logic for any handling of a flow removed message belongs in valve.py, rather than in the faucet.py shim).

1. A new PR that with just a flow removed handler, that calls a valve.py hook that just logs the message. That way we can test it on all our hardware platforms without any behavior change (and it would be nice to have this signal available).

2. Another new PR that adds some jitter around the existing hard timeout (meaning that relearning activity won't be synchronized as both you and Chris have pointed out). This would be a low risk change that involves no changes to the learning logic at all.

Once those 2 PRs are in, we can revisit the actual symptoms again to see what the most appropriate solution is.

anarkiwi:
Another possibility - when routing is enabled, there's periodic ARP/ND activity. Might be interesting to use that activity to refresh the learning rules as well (since the packet in "tax" has been paid may as well use it).

trungdtbk:
Hi Josh,
Sorry for the delay. I'm still working on it. My point is to eliminate hard_timeout.
I've implemented the following algorithm but it fails a test and I'm troubleshooting it.
Will try to fix this soon :)

Three timeouts are used:  idle_timeout_1 is for src_table, learn_timeout is for the controller and idle_timeout_2 is for dst_table;  idle_timeout_1 > learn_timeout > idle_timeout_2.
When we learns flowremoved event, the mark the host as sending_expired, otherwise it's active.
When the learn_timeout strikes, we re-install the src and dst rule if the host is active, otherwise remove the host from cache. The dst rule may still be there but will be removed by the switch.

The concern that needs to be verified if overwriting an active rule has any impact on packets?

I also thought about having controller probe hosts with ARP/ND but I think it's more expensive and requires the controller to know about host IPs.

trungdtbk:
I've updated the PR.

anarkiwi:
Thanks. Sorry please is it possible to move those small changes out to a
new PR? We can then work on breaking this PR up into smaller pieces (like
the add flow remove handler with log only like discussed).

trungdtbk:
Sure. I will work on that soon.

PR #498, opened 2017-04-10
Makefile for developers

shivarammysore:


gizmoguy:
Hi Shivaram,

Thanks for our PR.

A few points of feedback:

 * Remove sdist and clobber targets, packaging will be handled automatically by travis moving forward
 * Remove list target which has unknown license status
 * Add documentation to docs/ directory

Thanks.

shivarammysore:
Thanks @gizmoguy for the comments.  We can continue to build packages via travis, but, the way at least I (without speaking for others) test is to build packages, install and then test - hence `sdist` target.  "`clobber`" target cleans up built files - leave it back in original state.  I agree that removing `list` target is good.

gizmoguy:
I personally think the better way to install the package is "pip install [location]" or using setup.py.

I suspect sdists are only useful for package uploads.

Ultimately I'm trying to cut down on the number of ways we have to support faucet installs. Travis can be used for testing package builds as well.

shivarammysore:
Hi Brad,
I have updated the PR.  I agree with you that everyone should use our supported pypi faucet package for installation and reporting bugs.  But, the source code way of installation is required for testing features, etc and is use for developers.  Hence, I have included the same.  
Thanks

gizmoguy:
Thanks for the documentation work, I'm good with that.

I'm still going to vote we remove the sdist/clobber rules however. I think it complicates new developers' lives too much, and no developer needs to generate the sdist unless they want to be a package maintainer. All you are doing with this makefile is letting someone type "make sdist" over "./setup.py sdist". Where setup.py is the regular method package maintainers will be used to.

Also can you please rebase & squash out the "test" commit from this PR before merging.

gizmoguy:
Hi @shivarammysore 

I think it would be best to split a78a16e into it's own PR

shivarammysore:
Hi Brad,
Yep.  I have removed that commit.  Will submit separate PR for that.

Thanks

gizmoguy:
Hi Shivaram,

Looks better. Could you also update the description of commit 6fc7fb1 so that it isn't just "test".

Brad

shivarammysore:
Hi Brad,
Thanks for finding this. I have fixed the same.

/Shivaram

PR #499, opened 2017-04-14
re-learning using flowremoved event

trungdtbk:
I verified this on AT switches and it works fine. This is enabled by default. Can be disabled by setting flowremoved: False under dp config. We can reduce unnecessary packet-ins when hard_timeout is used.
This is to replace the PR #486

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -21,6 +21,7 @@ from acl import ACL
 
 import networkx
 
+MIN_TIMEOUT = 30 # seconds
 
 class DP(Conf):
     """Object to hold the configuration for a faucet controlled datapath."""
@@ -151,6 +152,7 @@ class DP(Conf):
             assert isinstance(port, Port)
         for acl in self.acls.values():
             assert isinstance(acl, ACL)
+        assert self.timeout >= MIN_TIMEOUT # pylint: disable=no-member
 
     def set_defaults(self):
         for key, value in self.defaults.items():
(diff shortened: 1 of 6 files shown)


anarkiwi:
Thanks. Please though, can you separate the jitter function into a separate
PR? I realize in my most recent request I was unclear about that. There is
some work required on the tests (for example, the hardcoded sleep(40)), the
jitter function first should be a small low risk change and easy to test.

trungdtbk:
Hi Josh, the PR #486 was updated to contain on jitter function.
This PR is for new active re-learning method that I propose. I set the limit to the minimum timeout users can configure to 30s. Therefore, the sleep(40) is to see if hosts disappear from the system.

anarkiwi:
Sorry for the confusion.

I'm looking for a PR that contains the learn jitter, and only the learn
jitter, first. It should be a small and simple change.

Then we can move forward with the change in the learning algorithm.

trungdtbk:
Hi Josh,
I updated the PR #486 to contain only the change regarding jitter. I changed its title to avoid confusion.

anarkiwi:
Thank you very much for this, and for your patience. I'll test in my setup
too, but I fully expect your change to be a great change for helping manage
learning load.

anarkiwi:
This PR will need rebasing for re-assessing. If you still see a problem with the latest code, let's revisit.

PR #527, opened 2017-05-02
a fix for permanent learning port fails to learn when reloading config

trungdtbk:
Resubmitting the PR #519

PR #532, opened 2017-05-02
a fix for permanent learning port fails to learn when reloading config

trungdtbk:
Reopen for #527. 
I could not replicate the error appeared on #531. But the reason could be the lack of delay for Faucet to finish the flow installation. I added time.sleep after the hup_faucet, hope this will fix it.


trungdtbk:
I should have thought of using "wait-for-flow".
I've updated the test.

anarkiwi:
Thank you, that fixed it.

PR #604, opened 2017-05-17
Prometheus MAC learning

Bairdo:
Exports the MAC learning 'tables' with Prometheus. 
E.g
```txt
learned_macs{dpid="0x3728fda4",n="0",port="2",vlan="100"} 68492399979869.0
```
68492399979869.0 = 3E:4B:21:2B:AD:5D
'n' is the 'dimension' to allow multiple MAC addresses on the one port.

https://travis-ci.org/Bairdo/faucet/builds/233057060

The test will add 4 hosts, check that they are exported, then wait 61 seconds for them to timeout, before adding another 2 hosts, where it will then check that the latest 2 are exported with the previous 4 either zeroed out or overwritten.
I have had problems with the test passing consistently when the timeout is set lower (30s or 40s).


anarkiwi:
Excellent thank you! May be a couple of days delay - at WAND right now
using their FAUCET deployment.

anarkiwi:
Do you have a Travis URL we can see that shows the tests passing?

anarkiwi:
Apologies, I missed it.

Bairdo:
[218598](https://github.com/REANNZ/faucet/pull/604/commits/218598cbc1401d1771c3f76122224dec2fd2e643)

I have run the test 30 times without failing once (20s base timeout). However it looks rather hacky, by doing multiple pings to keep the MAC addresses known to faucet.

When using a base timeout of 5s, 10s, & 15s the test fails.

There is also a small improvement made to compile the regex that is is used within a loop (but that should be unrelated to the long timeout issue)

I still need to remove the debugging print statements.


I will continue to investigate, maybe by making the pings simultaneous so the timeout values are closer together.

Bairdo:
[0dc5123](https://github.com/REANNZ/faucet/pull/604/commits/0dc51230c30950d8851a85f6b7a0d9758056455d)

https://travis-ci.org/Bairdo/faucet/builds/233865749


Changed approach to have first host **fping** the IP addresses of the second host.
Have run the test 30 times without failure.

Base timeout is set to 10s. 5s failed.

Bairdo:
[3f8e3af](https://github.com/REANNZ/faucet/pull/604/commits/3f8e3af9e7521f5484954be003f298b33c93bcd3)

https://travis-ci.org/Bairdo/faucet/builds/233874462

removes debug printing.

anarkiwi:
Thanks! Looks like some merge conflicts (due to addition of faucet_metrics.py). Almost there...

PR #756, opened 2017-07-14
Add directions for testing self-signed certificates

zhiweicen:
Self-signed certificates have the benefit of not relying on a CA but can still provide certain security of the control plane network. This document describes the steps needed to test if a switch supports self-signed certificates.

### DIFF
--- a/docs/self_signed_certs.md
+++ b/docs/self_signed_certs.md
@@ -1,0 +1,20 @@
+# Test Faucet TLS using self-signed certificates on the switch
+
+This document outlines the steps needed to test that a switch supports self-signed certificates for TLS based Openflow connections.
+
+## Prepare the keys and certificates.
+### Generate key pairs for the controller.
+    /usr/bin/openssl genrsa -out /tmp/ctrlr.key 2048
+    /usr/bin/openssl req -new -x509 -nodes -days 3650 -subj '/C=US/ST=CA/L=Mountain View/O=Faucet/OU=Faucet/CN=CTRLR_1' -key /tmp/ctrlr.key -out /tmp/ctrlr.cert
+### Generate key pairs for the switch.
+    /usr/bin/openssl genrsa -out /tmp/sw.key 2048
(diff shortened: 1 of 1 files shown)


shivarammysore:
This is good and very useful info.

Is storing certs under /tmp good?  /tmp is cleaned out at reboots.

Also, other things to consider:

   1. OVS stores certs under /var/lib/openvswitch and /etc/openvswitch
   directories.  Would storing these under those be good?
   2. When Controller certs are provided to Faucet, they must have the same
   permission of user who is running faucet service - normally, this could be
   "root" or "faucet".  If faucet, then chowm cert files to facuet:faucet (if
   faucet is user and faucet is group)
   3. One other thing to add is if we have more than 1 switch connecting
   via TLS to the same controller, how to add certs to the 2nd switch so that
   it can work with the already running controller.


Thanks

zhiweicen:
Thanks for the comments.

I've made another PR to make some improvements.

> Is storing certs under /tmp good? /tmp is cleaned out at reboots.
My original intention was to write the guidelines as instructions for
testing purpose. I realized somebody may want to run this for real.
Yes It's good to support those as well. I updated the paths in the PR.

> Also, other things to consider:
>
> 1. OVS stores certs under /var/lib/openvswitch and /etc/openvswitch
> directories. Would storing these under those be good?
Instead of using OVS folders I used /etc/ryu folder instead as this is
not strictly related to OVS.

> 2. When Controller certs are provided to Faucet, they must have the same
> permission of user who is running faucet service - normally, this could be
> "root" or "faucet". If faucet, then chowm cert files to facuet:faucet (if
> faucet is user and faucet is group)
Added a comment reminding the user doing that.

> 3. One other thing to add is if we have more than 1 switch connecting
> via TLS to the same controller, how to add certs to the 2nd switch so that
> it can work with the already running controller.
Added a section with the instructions to support that.

Zhiwei

shivarammysore:
Thanks Zhiwei.

PR #763, opened 2017-07-16
Eliminate hard-timeout, zero packet-ins

trungdtbk:
This is to replace #485.
By replacing hard-timeout with idle-timeout, we can miminize packet-ins to zero and keep the network working for all active hosts without Faucet running.
It is configurable.

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -1122,14 +1122,11 @@ class Valve(object):
         Expire state from the host manager only; the switch does its own flow
         expiry.
         """
-        ofmsgs = []
         if not self.dp.running:
             return
         now = time.time()
         for vlan in list(self.dp.vlans.values()):
-            ofmsgs.extend(self.host_manager.expire_hosts_from_vlan(vlan, now))
-
-        return ofmsgs
+            self.host_manager.expire_hosts_from_vlan(vlan, now)
 
     def _get_eth_srcs_learned_on_port(self, dp, port_no):
         old_eth_srcs = []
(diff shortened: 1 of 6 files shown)


trungdtbk:
Travis log: https://travis-ci.org/trungdtbk/faucet/builds/253409130

tests/faucet_mininet_test_unit.py
```diff
@@ -3297,3 +3297,61 @@ def test_switching(self):
...
+                native_vlan: 100
+                description: "b4"
+"""
+
+    def test_untagged(self):
+        self.net.pingAll()
+        first_host, second_host = self.net.hosts[:2]
+        third_host, fourth_host = self.net.hosts[2:]
+        first_host.cmd('ping %s &' % second_host.IP())
+        time.sleep(30) # wait enough for hosts to expire
```
KitL: Is there a way we can monitor the of channel so we can see the expiry messages, rather than just waiting for 30 seconds. That is a very long wait in the test suite..

faucet/valve.py
```diff
@@ -1361,6 +1366,23 @@ def get_config_dict(self):
             'acls': acls_dict,
             }
 
+    def flow_timeout(self, table_id, match):
+        eth_src = None
+        match_oxm_fields = match.to_jsondict()['OFPMatch']['oxm_fields']
+        if table_id == self.dp.eth_src_table:
```
KitL: What if the flow times out in the eth_dst table? Could you clarify for me how you are avoiding the rule in eth_src but not eth_dst table situation. 

Cause it's nice to reduce packet_ins but we **MUST NOT** allow ever having addresses time out in the eth_dst table before the eth_src. Unfortunately it is very hard for the test suite to cover that situation, if you can think of a way to test it I would appreciate that.

trungdtbk: The dst rules cannot expire before the src rules. If a host keeps sending without receiving, the src rule stays. The controller then refresh the dst rule for every "expire" period.
If the host may receive without sending, then the controller expires it as normal.

KitL: That does mean we have potentially 5 minutes of flooding right? It does seem plausible that on a large network, there would just be a constant number of hosts that you can expect to have all their traffic flooded at any given time.

When we could avoid this just by having a rule that catches when eth_dst rules time out and reinstalls them or removes the src matches to force them to be relearnt. Or we could have them never time out, and just have them be cleared every expire period, (if there is a sensible way to have the cache clear them, I am not sure off the top of my head what that would look like).

trungdtbk:
Hi Chris,
I updated the PR. The idea is if the controller sees:
- a flowremoved for a src rule, it just marks the host as expired
- a flowremoved for a dst rule, if the host is marked as expired, no action. otherwise, we re-learn the host assuming it is active.
the controller is not actively sending flowmods to refresh host rules.
the worst case scenario when the src rule is lost, the host will only obsolete for a period of timeout.
dst rule could timeouts before src rule, the controller immediately reinstalls dst rule upon receiving flowremoved for dst rule. It is not bad since the host is not actively receiving anything.
I just think this is an option that user can activate for their use.

KitL:
I like this implementation. It's ended up as something I think is pretty straightforward for anyone to understand.

I think it would be good to have a mechanism where the control plane cache checks it hasnt gotten out of sync with the data plane every so often. They definitely will get out of sync from time to time. I could imagine some weird behaviour creeping in around the edges of this implementation, but I cant come up with an example at the moment. Something like a conflict with some unusual combination of bad timing and some of the less likely configuration options..

I suggest that we bring this is still as non-default. I'll deploy it in the office and if it doesnt cause us any issues then we can try to move it to being the default down the line a bit.

anarkiwi:
Definitely worth further experimentation as non default - will you merge it?

KitL:
Ok it's running in our office. If you could set the default to False and tidy up the new conflict I'll merge this.

PR #776, opened 2017-07-23
Prometheus docker compose

KitL:
Updates docker-compose for use with prometheus

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,6 +1,6 @@
 version: '2'
 services:
-    db:
+    influxdb:
         restart: always
         image: 'influxdb:1.3.0'
         ports:
@@ -42,7 +42,7 @@ services:
         volumes:
             - '/opt/grafana:/var/lib/grafana'
         links:
-            - db
+            - influxdb
             - prometheus
 
     gauge:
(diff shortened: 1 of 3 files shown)


docker-compose.yaml
```diff
@@ -2,12 +2,12 @@ version: '2'
 services:
     db:
```
gizmoguy: I propose we rename the "db" container to "influxdb" now that we have multiple databases in play.

docker-compose.yaml
```diff
@@ -22,16 +22,28 @@ services:
...
     grafana:
         restart: always
-        image: 'grafana/grafana:3.1.1'
+        image: 'grafana/grafana:4.4.1'
         ports:
             - '3000:3000'
         volumes:
             - '/opt/grafana:/var/lib/grafana'
         links:
             - db
```
gizmoguy: I propose we rename the "db" container to "influxdb" now that we have multiple databases in play.

docker-compose.yaml
```diff
@@ -61,4 +73,6 @@ services:
             - '/etc/ryu/faucet:/etc/ryu/faucet'
         ports:
             - '6653:6653'
-            - '9244:9244'
+            - '9244'
+        links:
+            - db
```
gizmoguy: I propose we rename the "db" container to "influxdb" now that we have multiple databases in play.

PR #778, opened 2017-07-24
Fix pip install

### LINKED ISSUE #724: FAUCET pypi package doesn't install PBR
We have PBR set as in setup_requires however when we "pip install faucet" on a fresh machine we get something like this:

`pkg_resources.DistributionNotFound: The 'pbr>=1.9' distribution was not found and is required by the application`

Work around for now is to "pip install --upgrade pbr" before installing faucet.

Workout how to properly add PBR as a dependency so pip installs it.

reidrac:
Faucet requires pbr but is not included in the requirements.

Also the package in pypi is called 'faucet' and not 'ryu-faucet'

Fixes #724

### DIFF
--- a/docs/README_install.rst
+++ b/docs/README_install.rst
@@ -138,7 +138,7 @@ To install the latest pip package:
 .. code:: bash
 
   apt-get install python-dev
-  pip install ryu-faucet
+  pip install faucet
 
 To install the latest code from git, via pip:
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ influxdb
 ipaddress
 networkx
 numpy
+pbr
 prometheus_client
 pyyaml
 ryu==4.14


gizmoguy:
Thanks for the patch!

I had previously attempted this as a fix for #724 but it doesn't seem to work for me:

```
root@386804177f19:/faucet# python3 setup.py sdist
root@386804177f19:/faucet# pip3 install dist/faucet-1.5.4.tar.gz 
Processing dist/faucet-1.5.4.tar.gz
    Complete output from command python setup.py egg_info:
    
    Installed /tmp/pip-nokvq7ko-build/.eggs/pbr-1.5.4-py3.5.egg
      File "<string>", line 1, in <module>
      File "/tmp/pip-nokvq7ko-build/setup.py", line 63, in <module>
        pbr=True
      File "/usr/lib/python3.5/distutils/core.py", line 108, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 317, in __init__
        self.fetch_build_eggs(attrs['setup_requires'])
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 372, in fetch_build_eggs
        replace_conflicting=True,
      File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 849, in resolve
        raise DistributionNotFound(req, requirers)
    pkg_resources.DistributionNotFound: The 'pbr>=1.9' distribution was not found and is required by the application
```

I think PBR does some weirdness around including itself as a requirement that I don't fully understand yet. Did this fix work for you?

reidrac:
I think it did work, but now I'm no sure.

You need pbr installed to run sdist. Then the resulting distribution tarball should be fixed. Untar it on a system with no pbr (I used a virtualenv) and python `setup.py install` should work.

I'll confirm later when I'm back to the office.

reidrac:
You're right, setup.py requires pbr to install.

What confuses me is that I haven't seen that requirement when installing with pip pbr powered packages, so the problem must be somewhere else.

I'm closing this PR and I'll investigate further.

PR #798, opened 2017-08-02
Stricter checks to workaround possible Ryu packet parser hangs.

anarkiwi:


shivarammysore:
Hi Josh,
Confirming that this fix will address my issue without Ryu being patched.

I tested this with Ryu 4.16 which was released a few minutes ago and did a
pip install git+https://github.com/REANNZ/faucet.git

Thanks for this.

Regards

/Shivaram

anarkiwi:
Excellent, thanks for the confirmation.

We'll update to Ryu 4.16 in the next release.

PR #814, opened 2017-08-06
Updated to include discussion items from the mailing list

shivarammysore:
https://list.waikato.ac.nz/pipermail/faucet-dev/2017-August/000253.html

### DIFF
--- a/docs/architecture.rst
+++ b/docs/architecture.rst
@@ -8,7 +8,10 @@
 Faucet Design and Architecture
 ==============================
 
-Faucet enables providing L2 Switches for the masses not classes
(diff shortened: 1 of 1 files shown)


gizmoguy:
Hi Shivaram,

I think including umbrella here without a link to the paper that describes the technique will cause some confusion here.

If you want, maybe throw up a github issue to describe what paragraphs you want added to the architecture document and I'll write them up this week?

I see the language is already a little clunky and ambiguous in the document so it probably requires a few fixups so it makes sense.

Thanks

gizmoguy:
Issue #866 created to track this.

PR #867, opened 2017-08-25
assert sharing BGP speaker VLAN only for BGP-enabled vlans

trungdtbk:
This allows us to run BGP in one of VLANs with multi-vlan routing.

### DIFF
--- a/faucet/config_parser.py
+++ b/faucet/config_parser.py
@@ -113,7 +113,8 @@ def _dp_add_vlan(vid_dp, dp, vlan):
     if vlan not in dp.vlans:
         dp.add_vlan(vlan)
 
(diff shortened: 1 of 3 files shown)


trungdtbk:
Yes, it's a workaround for single BGP session support in Ryu. Ruy seems not to have any plan to make it multi-session, as they said no real application.
I also don't think it's necessary to have multiple speakers (one for each VLAN). Should we have BGP speaker that is agnostic to VLAN configuration?

anarkiwi:
I think we need to separate the issues of being able to have multiple speakers from multiple VLANs with speakers. Either way, it would be good to see a test that has a working config so that the workaround stays working and doesn't get broken.

PR #896, opened 2017-08-28
reduce learn timeout setting in Idle-timeout test to make it reliable.

trungdtbk:
I figured out the cause for unreliable tests. The test uses timeout 8secs; this plus with a jitter, sometimes making it to 13. The idle-timeout plus 2 for dst flow, hence the dst flows timeouts at 15 sec. The test waits for flowremoved at max 15 sec making the tests fails.
I reduced the learn timeout to 1sec in tests. that solved the problem.

### DIFF
--- a/tests/faucet_mininet_test_unit.py
+++ b/tests/faucet_mininet_test_unit.py
@@ -3844,7 +3844,7 @@ vlans:
         description: "untagged"
 """
     CONFIG = """
-        timeout: 8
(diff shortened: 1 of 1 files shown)


anarkiwi:
I'll merge this for now to continue the discussion but I'm not convinced that there is not a brief period where switching for the host is broken.
It would be good to test explicitly that switching is not broken.

PR #919, opened 2017-08-31
Add a git config to replace git url with https url

creneCSCO:
This configuration will allow to build the docker image when using proxy server. This is required because the
mininet/util/install.sh command is using a git url when pulling the openflow repository.

This extra line in the Dockerfile.tests allows to build the docker image using a command line like
sudo docker build --build-arg HTTP_PROXY=http://... --build-arg HTTPS_PROXY=http://... -t faucettests -f Dockerfile.tests .

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -44,6 +44,7 @@ RUN \
 ###
 
 RUN \
+  git config --global url.https://github.com/.insteadOf git://github.com/ && \
   /bin/echo -e "https://github.com/mininet/mininet\nhttps://github.com/google/pytype.git" |parallel --bar git clone
 
 RUN \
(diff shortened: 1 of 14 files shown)


anarkiwi:
Thanks - please would you be able to move the statement earlier (there are other git clones).

creneCSCO:
The previous git clone in the Dockerfile is ok because it is using https url.
The main issue is related to the git clone git:// which is done while executing the command "mininet/util/install.sh -nfv"

But i can still move the new command to the previous RUN section.

anarkiwi:
It'd be great to move it further up so that if we add further git config magic it will apply consistently.

anarkiwi:
Thanks! Merged.

PR #1086, opened 2017-10-11
Update docker containers used for docker-compose.

gizmoguy:
* Upgrade influxdb & update environment variables
* Upgrade prometheus
* Upgrade grafana


gizmoguy:
I think the configuration of influxdb is still wrong, investigating now.

Don't merge for now.

gizmoguy:
Ok I've fixed this PR! Working locally now

PR #1148, opened 2017-10-28
add compose file for rpi

cglewis:
Note: `couch` was removed, since I couldn't find an image that worked on `armhf`.  Everything else seems to work though, so I'm not sure if that matters for `gauge` or not?


anarkiwi:
couch considered experimental/demo quality, so fine to defer support for it. Thanks for the contribution!

PR #1203, opened 2017-11-15
Remove sw2 from config; fixing unhandled exception on startup

mab68:
Removed content under sw2 in the faucet.yaml config file, which was causing a crash when starting the docker.


gizmoguy:
Actually I had a better idea for how to fix this bug.

Add a third interface to sw2 and set it on the "guest" VLAN. This should allow the faucet config sanity checker to validate the configuration.

Issue #1201, opened 2017-11-15
Unhandled Exception when referencing a unconfigured ACL


mab68:
Steps to reproduce:
1. Use this config:
```
vlans:
    guest:
        vid: 200

dps:
    sw2:
        dp_id: 0x2
        hardware: "Allied-Telesis"
        interfaces:
            1:
                name: "pi"
                description: "Raspberry Pi"
                native_vlan: guest
                acl_in: access-port-protect
```
2. Start faucet
3. Crash:

```
Nov 15 01:59:34 faucet.exception ERROR    Unhandled exception, killing RYU
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/faucet/valve_util.py", line 34, in __koe
    func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/faucet/faucet.py", line 146, in _load_configs
    new_config_file, self.logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 50, in dp_parser
    config_hashes, dps = _config_parser_v2(config_file, logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 165, in _config_parser_v2
    top_confs['vlans'])
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 133, in _dp_parser_v2
    dp.finalize_config(dps)
  File "/usr/lib/python3.6/site-packages/faucet/dp.py", line 508, in finalize_config
    resolve_acls()
  File "/usr/lib/python3.6/site-packages/faucet/dp.py", line 476, in resolve_acls
    port.acl_in = self.acls[port.acl_in]
KeyError: 'access-port-protect'
```

gizmoguy:
Fixed in #1210

PR #1210, opened 2017-11-16
Fix referencing an unconfigured ACL #1201

mab68:
Fixed issue #1201.


anarkiwi:
This does appear to fix the problem, but we do need a test as well. test_check_config.py would be a good place to add it.

anarkiwi:
We're getting closer - but the test should check that the config checker fails (unconfigured ACL).
The parser can assert that the ACL exists, and the assert will be caught by the config parser and reported.

PR #1219, opened 2017-11-16
keep track of src_ip for learning hosts, if exists

cglewis:
if there is no `src_ip` it gets stored as `None`.  Currently only packets with a `protocol_name` of `ipv4`, `ipv6`, and `arp` are used to capture a `src_ip`.


anarkiwi:
Closing as will do different way today, as discussed.

PR #1230, opened 2017-11-19
port range configuration

trungdtbk:


anarkiwi:
This will be very useful!

Please add a test that shows overlapping changes are handled (interfaces from interfaces: and from range:).

Please also move the tests to the test_valve.py etc class. We are wanting to use that test suite more as it's much faster which is better for developers and it will enable our static analysis tools to work better.

trungdtbk:
@anarkiwi You mean the tests for valid/invalid configs should be moved to test_valve.py? 
The test in test_valve.py does check whether we can reload the config with new configuration from interface-ranges.
I will extend it to check whether new configuration from interface will override the interface-ranges.

KitL:
This should really go in a test_config.py, test_valve should assume config parsing has already been tested.

Mark is working on writing test_config.py as we speak. It should be up later today.

I think in its first state it will just be a clone of test_config_check.py, but we are gonna move that to actually testing the objects that are created rather than just whether or not it passes the config checking.

trungdtbk:
OK. I'll wait for Mark.

KitL:
The pull request is up now, I'm just waiting for the checks to pass.

KitL:
I'd like to have a test in test_config that verifies that the Port objects actually have the things you have configured. So something like:
```
_, dps = config_parser.dp_parser(f, log)
self.assertTrue(all(['drop_non_ospf_ipv4' == p.acl_in for p in dps[0].ports])) 
```
Then once we have verified that this config gives us the Conf objects that we want, there should be no need to test this in test_valve. test_valve should just test that given a sensible Conf the Valve object functions correctly.

trungdtbk:
agree with that. I will update.

Issue #1209, opened 2017-11-16
Unhandled exception when referencing an unconfigured vlan acls


mab68:
Steps to reproduce:
1. Use this config:
```
vlans:
    office:
        vid: 100
        description: "office network"
        acl_in: office-vlan-protect
        faucet_mac: "0e:00:00:00:10:01"
        faucet_vips: ['10.0.100.254/24', '2001:100::1/64', 'fe80::c00:00ff:fe00:1001/64']
        routes:
            - route:
                ip_dst: '192.168.0.0/24'
                ip_gw: '10.0.100.2'
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        proactive_learn: True
        interfaces:
            1:
                name: "h1"
                description: "host1 container"
                native_vlan: office
                acl_in: access-port-protect
```
2. Start faucet
3. Crash:
```
Nov 16 02:50:28 faucet.exception ERROR    Unhandled exception, killing RYU
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/faucet/valve_util.py", line 34, in __koe
    func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/faucet/faucet.py", line 146, in _load_configs
    new_config_file, self.logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 50, in dp_parser
    config_hashes, dps = _config_parser_v2(config_file, logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 165, in _config_parser_v2
    top_confs['vlans'])
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 133, in _dp_parser_v2
    dp.finalize_config(dps)
  File "/usr/lib/python3.6/site-packages/faucet/dp.py", line 509, in finalize_config
    resolve_acls()
  File "/usr/lib/python3.6/site-packages/faucet/dp.py", line 472, in resolve_acls
    vlan.acl_in = self.acls[vlan.acl_in]
KeyError: 'office-vlan-protect'
```

mab68:
Fixed in #1234

PR #1234, opened 2017-11-19
Fix referencing unconfigured vlan acls, issue #1209

mab68:


anarkiwi:
This is a check for unconfigured ACL, not VLAN?

mab68:
My bad, renamed appropriately.

Issue #1244, opened 2017-11-20
Unhandled exception when vlan vips are not valid strings


mab68:
Config:
```
include:
    - acls.yaml
vlans:
    office:
        vid: 100
        description: "office network"
        acl_in: office-vlan-protect
        faucet_mac: "0e:00:00:00:10:01"
        faucet_vips: ['aaaaa', '', '123421342']
        routes:
            - route:
                ip_dst: '192.168.0.0/24'
                ip_gw: '10.0.100.2'
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        proactive_learn: True
        interfaces:
            5:
                name: "trunk"
                description: "VLAN trunk to sw2"
                tagged_vlans: [office]
                acl_in: access-port-protect
```
Crash:
```
Nov 20 00:56:32 faucet.exception ERROR    Unhandled exception, killing RYU
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/faucet/valve_util.py", line 34, in __koe
    func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/faucet/faucet.py", line 197, in _load_configs
    new_config_file, self.logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 51, in dp_parser
    config_hashes, dps = _config_parser_v2(config_file, logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 168, in _config_parser_v2
    top_confs['vlans'])
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 121, in _dp_parser_v2
    vlans[vlan_ident] = VLAN(vlan_ident, dp_id, vlan_conf)
  File "/usr/lib/python3.6/site-packages/faucet/vlan.py", line 134, in __init__
    ipaddress.ip_interface(btos(ip)) for ip in self.faucet_vips]
  File "/usr/lib/python3.6/site-packages/faucet/vlan.py", line 134, in <listcomp>
    ipaddress.ip_interface(btos(ip)) for ip in self.faucet_vips]
  File "/usr/lib/python3.6/ipaddress.py", line 119, in ip_interface
    address)
ValueError: 'aaaaa' does not appear to be an IPv4 or IPv6 interface
```
faucet_vips not being a sequence of strings in #1240

mab68:
```
include:
    - acls.yaml
vlans:
    office:
        vid: 100
        description: "office network"
[...]

Issue #1241, opened 2017-11-20
Unhandled exception when route ip_dst or ip_gw is not an ipv4 or v6 address


mab68:
Config:
```
include:
    - acls.yaml
vlans:
    office:
        vid: 100
        description: "office network"
        acl_in: office-vlan-protect
        faucet_mac: "0e:00:00:00:10:01"
        faucet_vips: ['10.0.100.254/24', '2001:100::1/64', 'fe80::c00:00ff:fe00:1001/64']
        routes:
            - route:
                ip_dst: ''
                ip_gw: 'aaaaa'
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        proactive_learn: True
        interfaces:
            5:
                name: "trunk"
                description: "VLAN trunk to sw2"
                tagged_vlans: [office]
                acl_in: access-port-protect
```
Crash:
```
Nov 20 00:28:47 faucet.exception ERROR    Unhandled exception, killing RYU
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/faucet/valve_util.py", line 34, in __koe
    func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/faucet/faucet.py", line 197, in _load_configs
    new_config_file, self.logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 51, in dp_parser
    config_hashes, dps = _config_parser_v2(config_file, logname)
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 168, in _config_parser_v2
    top_confs['vlans'])
  File "/usr/lib/python3.6/site-packages/faucet/config_parser.py", line 121, in _dp_parser_v2
    vlans[vlan_ident] = VLAN(vlan_ident, dp_id, vlan_conf)
  File "/usr/lib/python3.6/site-packages/faucet/vlan.py", line 150, in __init__
    ip_gw = ipaddress.ip_address(btos(route['ip_gw']))
  File "/usr/lib/python3.6/ipaddress.py", line 54, in ip_address
    address)
ValueError: 'aaaa' does not appear to be an IPv4 or IPv6 address
```
Similar to #1240

PR #1262, opened 2017-11-21
Fix issue #1244 and clean up some test cases

mab68:
Fix unhandled exception with invalid strings #1244 and reduced clutter in some test cases.


mab68:
Also fixes #1241

anarkiwi:
It'd also be good for config parser failures, to raise a new exception created for the purpose - something like InvalidConfigError (so we aren't catching a litany of other random exceptions).

PR #1267, opened 2017-11-22
Created exception for invalid configs and created some tests to go with

mab68:
I'll update the other pull requests once this one goes through.


KitL:
I guess some of your other PRs introduce conflicts with this one. Sorry, didnt realise that was the case. You can close your pull requests if they arent needed any more.

Cheers,

PR #1343, opened 2017-12-04
Add fuzz testing for faucet.yaml config file

mab68:
Fuzz testing using py-afl-fuzz.


fuzz-requirements.txt
```diff
@@ -0,0 +1,17 @@
+bitstring
```
gizmoguy: Protip you can use "-r filename.txt" to include another requirements file.

I suspect the top of this file could be replaced with "-r test-requirements.txt"

mab68: AFL seems to take the test cases (provided in tests/in) and alters them in ways to produce errors. The mass amount of smaller files gives better results performance wise over one bigger file.

gizmoguy:
Great, thanks for the cleanup of fuzz-requirements.txt

Other thing I would suggest is moving the example configurations that seed the AFL fuzzer to a directory with a better name, perhaps "tests/fuzzer/config/"

gizmoguy:
This should be rebased/squashed now to clean up some of those commits.

We also need some documentation on how to run the fuzzer added to Sphinx.

PR #1348, opened 2017-12-04
Add documentation of config to Conf object source files

KitL:
It would be good for someone to do a basic editing check of this. Not sure how this will render with the readthedocs stuff but I figured it would be best to get it up and on git earlier rather than later.

### DIFF
--- a/faucet/acl.py
+++ b/faucet/acl.py
@@ -20,7 +20,35 @@ from faucet.conf import Conf
 
 
 class ACL(Conf):
-    """Implement FAUCET configuration for an ACL."""
+    """Contains the state for an ACL, including the configuration.
+
+ACL Config
+===========
+
+ACLs are configured under the 'acls' configuration block. The acls block
+contains a dictionary of individual acls each keyed by its name.
+
+Each acl contains a list of rules, a packet will have the first matching rule
+applied to it.
+
+Each rule is a dictionary containing the single key 'rule' with the value the
+matches and actions for the rule.
+
+The matches are key/values based on the ryu RESTFul API.
+The key 'actions' contains a dictionary with keys/values as follows:
+* allow (bool): if True allow the packet to continue through the Faucet
+    pipeline, if False drop the packet.
+* meter (str): meter to apply to the packet
+* output (dict): used to output a packet directly. details below.
+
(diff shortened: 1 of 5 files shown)


byllyfish:
The text looks good.

I did a quick build with Sphinx and there were two types of warnings.

  1. Sphinx doesn't like your section titles. (Unexpected section title.) I fixed this warning by removing them.
  2. When ending a line with a colon, insert a blank line. (Unexpected indentation.)

I also indented your bullet lists one space. 

With these fixes applied to the ACL file, the sphinx warnings went away.

```python

class ACL(Conf):
    """Contains the state for an ACL, including the configuration.

ACLs are configured under the 'acls' configuration block. The acls block
contains a dictionary of individual acls each keyed by its name.

Each acl contains a list of rules, a packet will have the first matching rule
applied to it.

Each rule is a dictionary containing the single key 'rule' with the value the
matches and actions for the rule.

The matches are key/values based on the ryu RESTFul API.
The key 'actions' contains a dictionary with keys/values as follows:

 * allow (bool): if True allow the packet to continue through the Faucet 
   pipeline, if False drop the packet.
 * meter (str): meter to apply to the packet
 * output (dict): used to output a packet directly. details below.

The output action contains a dictionary with the following elements:

 * port (int or string): the port to output the packet to
 * swap_vid (int): rewrite the vlan vid of the packet when outputting
 * failover (dict): Output with a failover port. The following elements can be
   configured:

    * group_id (int): the ofp group id to use for the group
    * ports (list): a list of the ports the packet can be output through
"""

```

gizmoguy:
@KitL for future reference I've left documentation here on how to locally build the Sphinx documentation offline:

https://github.com/faucetsdn/faucet/blob/master/docs/README.rst

byllyfish:
Sphinx has its own Makefile (to make html). Here's how I installed sphinx on my mac (which has Python3.6; other python versions will work), and then ran it.

```
git clone -b docs --single-branch https://github.com/KitL/faucet.git
cd faucet
python3.6 -m venv venv36
source venv36/bin/activate
pip install sphinx sphinx_rtd_theme
cd docs
make html
```

If it's easier, you can merge your edits and I will fix the sphinx warnings later.

Issue #1312, opened 2017-11-29
FAUCET event notification


anarkiwi:
@cglewis @corelight - FYI.

FAUCET knows about topology and network state changes - e.g. a host was learned, a port went up or down. We want a way to send these events to an external system, but we also can't have FAUCET block on sending the event (since that would affect the network).

We propose to provide an adaptor process that streams JSON'd events (first to RabbitMQ, later to the new Bro Broker). Current proposal is FAUCET will stream the events over a Unix domain socket to the adaptor process (fire/forget). We don't guarantee all events (FAUCET may be restarted, etc) will be streamed. Ryu hijacks thread/scheduling/sockets which FAUCET-side must be careful of.

This feature would not replace Prometheus in any way (the more general purpose NMS interface for FAUCET).

In the future an gNMI adaptor would be provided also.

Notes from Bro follow on the Broker implementation.
==

to use the new Broker version, there are branches called
topic/actor-system in both the main Bro repository
(git://git.bro.org/bro) and the Broker repository
(git://git.bro.org/broker); if you start with the Bro branch, it'll
check out the Broker submodule in the correct version.

Compiling Broker requires CAF 0.15.x from
https://github.com/actor-framework/actor-framework (0.15.4 minimum)

The Broker branch comes with updated documentation describing the new
APIs for C++ and Python: "make doc", then look into
build/doc/html/index.html (building the docs requires Sphinx, rst2pdf,
and "pip install sphinx_bootstrap_theme"). Look at "2.3. Exchanging
Bro Events" in particular. Broker's unit tests in tests/ demonstrate
further how the API works.

On the Bro side, you find the Broker API in
scripts/base/frameworks/broker/main.bro. Tests showing how to use it
are in testing/btest/broker/. Broker's C++ API maps pretty much
directly over to Bro. We are working on some more higher-level
communication patterns, too, but that's probably not that important
for you if it's only about sending events into Bro.

cglewis:
Perfect, sounds like a good approach to me. I think fire/forget as you mention is a good compromise here.  Happy to test RMQ on my end when the time comes.

shivarammysore:
Josh,
Well articulated.

Instead of sending events via Unix sockets - which means another port open
(need to guard it from security perspective), how about sending event
related data to Prometheus itself.  We can configure Alert Manager in
Prometheus to consume them appropriately.  It is my opinion that sending
event information itself will not be sufficient - there will be a followup
[...]

PR #1371, opened 2017-12-07
 adds a rabbitmq adapter to get events from the socket

cglewis:
- updates the FAUCET_EVENT_SOCK to be a file that lives in
/var/run so that volume mounts can be set for docker containers

- intended to be used with docker-compose by just adding additional yaml files to the command

- add tests and docs for the specific adapter

- add docs for more vendored adapters in the future

- if using a given adapter additional environment variables are used and prefixed with FA for FAUCET Adapter

see issue #1312


cglewis:
@gizmoguy Since you can't make a Docker volume that can share between containers as just a single file, I am putting into a dedicated folder (`/var/run/faucet/faucet.sock`), so that only that `faucet.sock` file inside the `faucet` dir is shared.  I also made `FAUCET_EVENT_SOCK` a bool to simplify that.  Let me know if that looks better.

cglewis:
looks like the markdown doesn't directly translate to rst. I'll fix that up and push up

gizmoguy:
@cglewis perfect! looks great now, thanks for those tweaks.

gizmoguy:
@cglewis I've kicked off a docker build for the new Dockerfile, hopefully it will be up here soon:

https://hub.docker.com/r/faucet/faucet-event-adapter-rabbitmq/

cglewis:
sweet!

PR #1375, opened 2017-12-07
FAUCET_EVENT_SOCK can be a bool (default path) or an explicit path.

anarkiwi:


cglewis:
@anarkiwi Good call. That fixed things, thanks. I'll be more diligent in future PRs.

PR #1377, opened 2017-12-08
Faucet testing with OVS on hardware

shivarammysore:


shivarammysore:
@gizmoguy - thanks for the review.

1. Router/Switch is included so that testers can understand the relationship between the test host and the switch to be tested.  IP Address name space and resolution.  We have seen questions in past on the mailing list and this should help understand the relationship better.
2. This is a RST version of https://docs.google.com/document/d/19xaSi3IKrz9buQN6F80YAYfJ9Fgy9Nbh5eV7_be3IYI/edit?usp=sharing - also has a the original diagram

In terms of RST changes, 

1. I tried to use the RST tables and it does not support the complexity of the table cells information that I have (early commit versions had the list-table but, it does not work in this case).  Hence, forced to use html table.
2. I have the PR ready to pushed for docs/vendors/index.rst - will send it out.
3. Good suggestion on removing comments from hw_switch_config.yaml for clarity - PR ready to be sent out.

gizmoguy:
Diagram:
1. Ok, we call this the Control Plane Network (or CPN) to signify that OpenFlow traffic goes over these links, I'll redraw your diagram for you now that I have the original google docs drawing.

RST:
1. Here is how your HTML table renders with the RTD template: https://i.imgur.com/1DSBG8c.png, it is currently unreadable so you'll need to find a new way to present your information

shivarammysore:
Ok Thanks.
I am thinking I will remove table completely, but use the same details in a
paragraph format.

shivarammysore:
Removed embedded html table completely.

gizmoguy:
I have made a new diagram for you - https://docs.google.com/drawings/d/1Of_eOlkO7uk9TxYGCDqp_RbuQos3tqt4mt57w_A8LJc/edit?usp=sharing

Some more changes for the RST:
1. Your "commands on:" sections are messed up, the syntax highlighter thinks all your commands are comments. Also the output for "ip a" has broken HTML entities inside it, do we really need a full output from "ip address show" here by the way? There are a bunch of DOWN interfaces that aren't used in testing that could be removed.

2. There is a bit of duplication in the document that we can clean up, for instance there is documentation for ethtool blinking port LEDs twice in the article.

3. Some minor language/flow issues but I can clean these up in a future PR once this one is merged in.

shivarammysore:
I fixed the "Commands on:" section.  When I converted RST to html, on my
machine, I had no errors, but, these are not straight forward.  Thanks for
reporting it.  Hope it fixes the issue now.  The PR should be updated now.
The duplication is intentional.  I really wanted to have it named on host
and switch side to make it easy for a newbie to understand the flow.  Yes -
one could optimize.
The way the diagram is done, it matches the output, configuration file,
etc.  I have take lots of pain to make sure they are all coordinated.  This
is from my experience of helping others to resolve issues.  Hope we can
maintain the same diagram format.

Thanks @gizmoguy for your review.

gizmoguy:
"commands on" section still looks broken to me, commands appear to be grayed out due to the syntax highlighter interpreting them as comments: https://i.imgur.com/ktDy6DG.png

For the duplication you could perhaps consider using the built in [inline referencing](http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role) feature of RST to continually link to the "debug" section instead of repeating in-text. This will help us if we ever need to update the documentation to ensure commands only need to be updated in one place.

Ok so regarding the diagram, my issue is that your setup as described in this document is going to be near impossible for anyone but yourself to implement, as you've used predictable network interface names which are tied to physical location of where the NIC is connected (onboard, USB, PCI-e riser, etc) and you haven't given us any detail on how your machine is configured. I attempted to simplify this and make it more reproducable by not using predicable network interface names.

I hope you appreciate our documentation in the past has been a little messy and all over the place and with this move to Sphinx/readTheDocs I am attempting to make sure our documentation is first class to enable new users the easiest transition into faucet, apologies if I seem pedantic but this is my current goal.

gizmoguy:
Ok I've noticed a lot of our bash CLI examples in the faucet documentation has the exact same issue as yours (grayed out commands), I will be working on a PR to fix this right now.

But just to help you out, the secret is to use "console" as your syntax highlighting mode instead of "bash".

shivarammysore:
Thanks Brad.

Console is used and updated - great suggestion.  It was not in the standard
RST document.  But, it looks like it is working.

Good suggestion  - I will update on document on inline_referencing.

In regards to diagram, I wanted to match the commands.  I wanted users to
understand port numbers and corresponding output so that they can easily
understand and co-relate.
Thanks

gizmoguy:
I would suggest we rewrite your command output to match the new diagram I drew - best of both worlds.

shivarammysore:
Updated.

gizmoguy:
Great! We're getting real close now.

I've made some minor positioning tweaks to the diagram, if you could include those that would be fantastic.

Looks like our syntax highlighting still isn't working for your shell commands, replace "#" with "$" and add sudo to the front of commands, this should fix the problem.

Then remove unused interfaces from the output of "ip address show", such as docker0, ens802f0, etc.

Finally, this PR has a messy git history now so if you could rebase/squash back onto faucet@master that would make my life easier.

Thanks.

shivarammysore:
@gizmoguy - thanks for the review.

All issues addressed and a new PR sent.

Thanks

PR #1380, opened 2017-12-10
Faucet Testing with OVS on hardware

shivarammysore:


shivarammysore:
Thanks Brad.  Fixed it.

PR #1396, opened 2017-12-12
Faucet and Gauge now inherit from a base image which uses su-exec instead of running as root

cglewis:
- updates to latest DPDK for OVS
- updates Dockerfiles to use base images that use `su-exec` where possible
- bit of cleanup on Dockerfiles
- base image defaults to user/group of 9001 unless specified as docker runtime by the user to override it.
- `su-exec` is used as the `ENTRYPOINT` and `CMD` gets passed to it so that actual process isn't run as root.  this means that if `ENTRYPOINT` is overridden in child Dockerfiles, it will no longer use the `su-exec` behavior.
- since Faucet and Gauge inherited from a public Python image, and the base image which runs `su-exec` has to be the parent, there's a new intermediate Python image, which is also updated (old image was over 2 years old)
- both the base image and the python image was expected to live publicly on the Docker Hub as `faucet/faucet-base:latest` and `faucet/faucet-python3:latest` respectively. (There are also the rPi counterparts: `faucet/faucet-base-pi:latest` and `faucet/faucet-python3-pi:latest`)
- Travis tests container has been updated with IPv6 flag
- Documentation has been updated for Faucet and Gauge Docker usage regarding the user id

I tested this on OSX, Linux, and ARM.  Note since the UID is so high, it shouldn't conflict with existing users and volume permissions should work like any other non-root user would, so plan accordingly for writable directories on your host filesystem where volumes are used.  Travis tests passed on my fork, but this probably could use some more testing by folks before merging just to make sure there aren't any unexpected consequences.


gizmoguy:
Hi @cglewis 

Thanks for this! I've been meaning to tidy up our dockers for a while new and it looks like you've made some good progress on this.

Since there's quite a bit of work here, let me spend a bit of time testing everything and I'll get back to you for review shortly.

cglewis:
@gizmoguy updated - let me know if that looks better or needs more tweaking. thanks!

docs/docker.rst
```diff
@@ -40,6 +34,9 @@ tag will always point to the latest git commit. All tagged versions of Faucet
 in git are also available to use, for example using the faucet/faucet:v1_3
 Docker will run the stable version 1.3 of Faucet.
 
+By default the Faucet and Gauge images are run as the `faucet` user under UID 0,
```
gizmoguy: I think this documentation is wrong, we should use ``-e LOCAL_USER_ID`` right?

-u doesn't work

> brad@nix/50:~/github/faucet$ docker run -it -u 100000 --rm -v /scratch/faucet_configs/:/etc/ryu/faucet/ faucet/faucet  
> Starting with UID : 0
> su-exec: setgroups(0): Operation not permitted

-e works:

> brad@nix/50:~/github/faucet$ sudo docker run -it -e LOCAL_USER_ID=100000 --rm -v /scratch/faucet_configs/:/etc/ryu/faucet/  faucet/faucet
> Starting with UID : 100000
> loading app faucet.faucet

cglewis: ah, yes. my bad. corrected.

gizmoguy:
I'm not entirely sure why yet - but your tweaks to Dockerfile.tests seem to make pytype die:

> brad@nix/50:/tmp/faucet$ sudo docker run --sysctl net.ipv6.conf.all.disable_ipv6=0 --privileged -ti faucet/tests
> ...
> File "check_faucet_config.py", line 21, in <module>: Couldn't import pyi for 'logging' [pyi-error]
>     File: "/usr/local/lib/python2.7/dist-packages/pytype/typeshed/stdlib/2/types.pyi", line 22
>       StringTypes = (StringType, UnicodeType)
> 
>   ParseError: syntax error, unexpected ',', expecting OR or AND or ')'
> File "check_faucet_config.py", line 22, in <module>: Couldn't import pyi for 'sys' [pyi-error]
>     File: "/usr/local/lib/python2.7/dist-packages/pytype/typeshed/stdlib/2/types.pyi", line 22
>       StringTypes = (StringType, UnicodeType)

I'm having a poke now to see if I can find the root cause

gizmoguy:
Ok a different version of pytype seems to be getting installed somehow.

Line 22 on /usr/local/lib/python2.7/dist-packages/pytype/typeshed/stdlib/2/types.pyi differs between my working docker and the new docker:

StringTypes = ...  # type: Tuple[Type[StringType], Type[UnicodeType]]
StringTypes = (StringType, UnicodeType)

gizmoguy:
Ok nevermind this has nothing to do with your changes - pytype updated the version of typeshed they use 6 days ago:

https://github.com/google/pytype/commit/a86093f79e5b63a081339ed814e50c4da27459c5

Obviously my working test dockers still have the old version. I will work on this tomorrow and see if I can make new typeshed work with faucet.

gizmoguy:
Ok I found the root issue! Have lodged an upstream bug with pytype:

https://github.com/google/pytype/issues/69

cglewis:
Would it be worth locking the commit for pytype to a known good until it's resolved?
```
./docker/base/Dockerfile.base-tests:    git clone https://github.com/google/pytype.git && \
./docker/base/Dockerfile.base-tests:    git checkout 81f77b17eca63ee718e1c57e533440db412bf847 && \
```
(not sure that's the exact commit you want, but just as an example)

gizmoguy:
Hi @cglewis 

Yes I think this a good idea, I'm quite keen to get this merged in today.

I found commit pytype@0964db4d6683324035486109a5ea6c210d8b131c worked, here's the patch I used:

```
diff --git a/docker/base/Dockerfile.base-tests b/docker/base/Dockerfile.base-tests
index 0eb403d..0681065 100644
--- a/docker/base/Dockerfile.base-tests
+++ b/docker/base/Dockerfile.base-tests
@@ -90,6 +90,7 @@ RUN \
     cd .. && \
     git clone https://github.com/google/pytype.git && \
       cd pytype && \
+      git checkout 0964db4d6683324035486109a5ea6c210d8b131c && \
       git submodule init && \
       git submodule update && \
       python2 $SETUPQ install && \
```

cglewis:
@gizmoguy pushed up

gizmoguy:
@cglewis thanks! I think we're getting close...

pytype have fixed their repo so we can roll that change back now.

I've also tweaked your entrypoint.sh, it now supports setting GID and turns out we do need to fix permissions on a few faucet directories so that faucet will start nicely:

```
#!/bin/bash

# Add local user
# Either use the LOCAL_USER_ID if passed in at runtime or
# fallback to 0 to maintain backwards compatibility

USER_ID=${LOCAL_USER_ID:-0}
GROUP_ID=${LOCAL_GROUP_ID:-0}

echo "Starting with UID=$USER_ID GID=$GROUP_ID"
export HOME=/home/faucet
addgroup -g $GROUP_ID faucet >/dev/null 2>&1 || true
adduser -u $USER_ID -G faucet -g "" -h $HOME -D faucet >/dev/null 2>&1 || true

# Update permissions on directories installed by pip before we su-exec
chown faucet:faucet /var/log/ryu/faucet/
chown -R faucet:faucet /etc/ryu/faucet/

exec /sbin/su-exec faucet "$@"
```

I still have a few more images to verify still boot up, but after that's done I'll get this all merged in!

cglewis:
cool, nice additions - pushed.

gizmoguy:
AH sorry, more changes. We will need to create `/var/run/faucet/` in entrypoint and own it to the faucet user.

cglewis:
no worries. how's that look?

gizmoguy:
@cglewis I've tested all the dockers and they look good to go, I decided that potentially overriding permissions on the host volume was bad so I've reverted my changes to entrypoint.sh:

```
#!/bin/bash

# Add local user
# Either use the LOCAL_USER_ID if passed in at runtime or
# fallback to 0 to maintain backwards compatibility

USER_ID=${LOCAL_USER_ID:-0}
GROUP_ID=${LOCAL_GROUP_ID:-0}

echo "Starting with UID=$USER_ID GID=$GROUP_ID"
export HOME=/home/faucet
addgroup -g $GROUP_ID faucet >/dev/null 2>&1 || true
adduser -u $USER_ID -G faucet -g "" -h $HOME -D faucet >/dev/null 2>&1 || true

exec /sbin/su-exec faucet "$@"
```

Just make sure to copy the whole thing into your repo (you missed the addgroup command last time) and when you've pushed that up to your repo I'll merge.

cglewis:
@gizmoguy sorry about that, should be correct now! thanks for working through this with me!

cglewis:
As an aside, maybe the volume permissions/UID/GID is something we aim for changing the default of in 2.0.0

gizmoguy:
@cglewis I'm thinking of adding documentation for now on how to run as a non-root user and keeping UID 0 the default - this is definitely something we could change in faucet2 and flip the default to non-root and allow the user to override and run as root.

As a side note - we are having monthly developer sync ups for planning faucet2 - would you like to be invited to the next one to propose some more changes and hear what we have planned?

cglewis:
@gizmoguy absolutely - feel free to reach out to me :)

PR #1446, opened 2017-12-22
Fix issue 1421: Decode port change reasons.

aappachi:


anarkiwi:
It would be good to consult a dict and catch the keyerror for unknown instead, which would be more efficient and readable.

aappachi:
Thanks Josh. Comments have been addressed. Instead of try/catch blocks, used the get method with default value option to make it further simple.

PR #1551, opened 2018-01-23
Stack neighbour discovery

Dualis:
Active verification of link status between switches in a stack.

The process is as follows:
1. Timed event triggers
2. Controller instructs switch to send a message on each stack port
3. Timeout thread scheduled
4. OpenFlow rules on adjacent switches reply to messages
5. Replies are parsed by the controller
6. Timeout thread triggers, disables ports that did not reply 

This implementation avoids controller to controller messaging, with the following consequences:
 ++ Turnaround time and jitter are reduced.
 ++ CPU and bandwidth use are reduced.
 -- Does not determine whether an adjacent switch has an active controller.
 -- Constraints on DPID and port numbers.

**Implementation details:**
Each (switch, port) pair is assigned a pseudo-MAC address, see valve._get_mac_for_port().
The controller builds a packet as so:
- src_mac = Pseudo-MAC for the local port
- dst_mac = Pseudo-MAC for the destination port
- eth_type = 0xFCE7
- payload = Time when the packet was created.

Each switch has a flow for each port, that matches exactly these packets and sends them back.
The reply packets miss the local rules, and go to the controller.

Packing the DPID and port number into a MAC address places limitations on the length of these values.
Currently they are 4 bytes for a DPID, and 1 byte for the port number.
If these are too small, this can be updated to piggyback off IPv6 instead

**Detecting controller failure:**
If we need to do this, it's a small extension to have the local port flows timeout on a short delay.

**Testing**
- [+] Manual testing with Mininet
- [ ] Hardware testing (Planned for tomorrow)
- [ ] Faucet tests

**Feedback appreciated**


anarkiwi:
This is a fairly significant change, and it has no tests.

Please would you break it into smaller pieces and include tests. 

It would be good too, if you can please a rough summary of what you're implementing to faucet-dev beforehand, so that your design can be understood clearly and we can catch any issues early.

anarkiwi:
There are some other issues here too - the synthetic eth_src/dsts might conflict with real hosts on the network, and the payload could be spoofed which would lead to undefined behavior.

It would be good to come back to the design stage and for you to propose and us to discuss the idea before proceeding down the implementation path.

Dualis:
OK, understood.
I'll work on understanding the testing framework, and assemble some tests for this.

And I'll get an email to dev-faucet later this week.

Cheers, feel free to close this pull request if you don't want it hanging around.

Issue #1547, opened 2018-01-22
mirror interface should support more than one port


anarkiwi:
Right now interface monitoring supports just one interface:

interfaces:
   mirrored_interface:
      number: 1
      native_vlan: 100
   mirroring_interface:
      number: 2
      mirror: mirrored_interface

@cglewis  It'd be useful for mirror to support a list so it can mirror potentially lots of interfaces. 

It'd also be useful to be able to mirror an entire vlan/list of vlans.

cglewis:
being able to mirror at L3 would be useful too.  perhaps some combination with ACLs?  but yes, for now just supporting multiple ports being mirrored to a port would be a great start.

PR #1590, opened 2018-01-31
Fix for issue #1547

nfz1:
Added support for multi-port mirroring. Config file now supports either a single mirror port entry (legacy) or a list of ports to mirror from, as mentioned in issue #1547.

Added two unit tests that check port mirroring works as expected:
**FaucetUntaggedMultiMirrorTest** will check that we can mirror from _multiple ports to a single port_.
**FaucetUntaggedMultiMirrorSepTest** will check that we can mirror from _a single port to multiple ports_.


grafnu:
Also need to update docs/configuration.rst -- I didn't see that in the pull request...

nfz1:
I updated the configuration docs and have also found a bug with my changes when using port names instead of numbers for mirroring (the problem has been fixed).

While testing the above fix I have found a bug which seems to be present in the current master. If we use port names instead of port numbers Faucet rejects the port config with an error (issue #1594).

Issue #1594, opened 2018-02-01
Using port names instead of numbers for mirroring causes faucet to reject port config


nfz1:
It seems that using a port name instead of a port number causes faucet to raise an error and reject the port configuration. Here is an example config:

```yaml
vlans:
    office:
        vid: 100
        description: "office network"
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        proactive_learn: True
        interfaces:
            1:
                name: "a1"
                description: "host1 container"
                native_vlan: office
            2:
                name: "b1"
                description: "host2 container"
                native_vlan: office
            3:
                name: "c1"
                description: "host 3 container"
                mirror: "a1"
```

The above config produces the following log output:

```
Feb 01 19:26:02 faucet ERROR    New config bad (Port 3 must have a VLAN, be a stack port, or have output_only: True) - rejecting
Feb 01 19:26:03 faucet ERROR    features_handler: unknown datapath DPID 1 (0x1)
Feb 01 19:26:04 faucet ERROR    features_handler: unknown datapath DPID 1 (0x1)
```

If we modify the config and use _port number 1_ instead of _the port name a1_ for mirroring, everything works fine.

```
Feb 01 19:33:13 faucet INFO     Add new datapath DPID 1 (0x1)
Feb 01 19:33:14 faucet INFO     DPID 1 (0x1) connected
Feb 01 19:33:14 faucet.valve INFO     DPID 1 (0x1) Cold start configuring DP
Feb 01 19:33:14 faucet.valve INFO     DPID 1 (0x1) Configuring VLAN office vid:100 ports:Port 1,Port 2
Feb 01 19:33:14 faucet.valve INFO     DPID 1 (0x1) Port 1 configured
Feb 01 19:33:14 faucet.valve INFO     DPID 1 (0x1) Port 2 configured
Feb 01 19:33:14 faucet.valve INFO     DPID 1 (0x1) Port 3 configured
```

Error occurs with current stable master as of Jan 30 (5f70815514).

nfz1:
The config specifies a port we want to mirror from. In code we translate this to a port we mirror to, which was inverted for the port by name translation code path. Eventually, the **output_only** flag was being set on the wrong port (we were setting the mirror from as output only, not the mirror destination port) causing faucet to later reject the config when performing the [finalize port method check](https://github.com/faucetsdn/faucet/blob/master/faucet/port.py#L193).

anarkiwi:
[...]

PR #1601, opened 2018-02-02
Fix error when using port names for port mirroring

nfz1:
Fixes issue #1594.


nfz1:
The issue was already resolved by 6aad7b4.

PR #1686, opened 2018-02-22
Minor docs update for missing table header

grafnu:


docs/.gitignore
```diff
@@ -1 +1,2 @@
 _build
+source
```
gizmoguy: Could we change this from "source" to "source/apidoc". Just so we don't accidentally hide changes in the source root directory.

PR #1688, opened 2018-02-23
rabbitmq adapter use FAUCET_EVENT_SOCK as path.

Bairdo:
Also added a couple of debug strings.

### DIFF
--- a/adapters/vendors/rabbitmq/rabbit.py
+++ b/adapters/vendors/rabbitmq/rabbit.py
@@ -27,6 +27,24 @@ import sys
 import pika
 
 
+def get_sys_prefix():
+    """This was copied from faucet.valve_util.
+    Returns an additional prefix for log and configuration files when used in
+    a virtual environment"""
+
+    # Find the appropriate prefix for config and log file default locations
+    # in case Faucet is run in a virtual environment. virtualenv marks the
+    # original path in sys.real_prefix. If this value exists, and is
+    # different from sys.prefix, then we are most likely running in a
+    # virtualenv. Also check for Py3.3+ pyvenv.
+    sysprefix = ''
+    if (getattr(sys, 'real_prefix', sys.prefix) != sys.prefix or
+            getattr(sys, 'base_prefix', sys.prefix) != sys.prefix):
+        sysprefix = sys.prefix
+
+    return sysprefix
+
+
 class RabbitAdapter:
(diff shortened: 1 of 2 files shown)


adapters/vendors/rabbitmq/rabbit.py
```diff
@@ -141,6 +140,7 @@ def main(self):
                     print("Unable to send event to RabbitMQ because: %s" % err)
                     print("The following event will be retried: %r" % buffer)
                     self.rabbit_conn()
+                sys.stdout.flush()
```
Bairdo: At somestage (during setting the adapter up the first time) the program would crash (typically when variables are set incorrectly iirc) but it wouldn't print any info.
So this could be removed if necessary.

cglewis:
@Bairdo won't this break backwards compatibility? I think you can also have `FAUCET_EVENT_SOCK` be a bool for FAUCET per this:

https://github.com/faucetsdn/faucet/blob/32e46e83b50ae0540ff10e05b26865739cd53186/faucet/valve_util.py#L96

Bairdo:
@cglewis let me check again my tests again, thought faucet was reading the variable as path

cglewis:
I think it can do either, and defaults to `/var/run/faucet/faucet.sock` if set to `1`.  But yeah, let's go off the tests, I might be misremembering.

Bairdo:
With https://github.com/faucetsdn/faucet/blob/32e46e83b50ae0540ff10e05b26865739cd53186/faucet/valve_util.py#L96


Just to clarify:
if var is 0: return 0
if var is 1: return _prefix+'/var/...'
if its a string return the string.

So I'll update the pr to expect 0, 1, or path.

cglewis:
cool, thanks for improving that!

cglewis:
ah, good point. defer to the maintainers on that one.

gizmoguy:
I think @byllyfish tidied up `FAUCET_EVENT_SOCK` in the past in PR#1422, will wait to get his :+1: on these changes.

And sorry I don't follow the conversation about `FAUCET_EVENT_SOCK` and the system prefix, what do you mean?

byllyfish:
+1 Looks okay to me.

rabbit.py is a self-contained script with one dependency (pika). If all you need is the get_sys_prefix function from valve_util, I would copy/paste the function rather than add Faucet as a dependency. I don't think get_sys_prefix should be part of the public Faucet api.

PR #1820, opened 2018-03-27
Refactor mininet test runner into reusable component

grafnu:
This is a functionality-neutral refactoring that makes it possible to use the faucet_mininet test foundation against any module of unit tests. This paves the way for things like breaking up the _unit.py file into smaller chunks, or adding extended bits of functionality that are run/manged separately, or used by an external faucet client derivative of some kind. "clib" stands for "client library", since the bigger intent is to have it be a reusable module for multiple test clients.


anarkiwi:
Was there a git mv intended? I'm trying to determine if this PR is about moving and renaming, or new code. Moving and renaming makes the most sense to start.

grafnu:
It was supposed to be moving and renaming... maybe I didn't do the git
magic appropriately -- let me take a look at it and get back to you...

grafnu:
Ok, I had to split it into two commits to keep it sane. If you look at each of the individual commits it's pretty easy to see what's going on. Looking at the changed set of files is crazymaking. PTAL.

PR #1831, opened 2018-04-05
Refactor tcpdump_helper into more flexible standalone class

grafnu:


grafnu:
Created #1849 -- will resolve that first and then come back to this one!

PR #1832, opened 2018-04-06
Lldp fix

KitL:
lldp wasnt working when you used interface_ranges because it would recast values from their default types then re-check whether they were the correct type or not.

Fix by accepting both types.


KitL:
Dont know what to make of this. As far as I can the code is all tested..

PR #1838, opened 2018-04-09
install library needed for pyyaml cloader.

Bairdo:


Bairdo:
The library is libyaml on debain and is already installed in the tests dockerfile

PR #1844, opened 2018-04-13
Hostapd

Bairdo:
This is copied from [gasket](https://github.com/bairdo/gasket).

- Dockerfile.hostapd should probably inherit from alpine.

- The docker-compose file expects hostapd to be connected to OVS, which is done after the container has started. When this is not the case change the interface in wired.conf and change eth3 to the appropriate interface.
changing MAC address can also be removed (if the the docker interface's MAC is known).

- ping is required so faucet will learn where hostapd is.

- docker-compose-gasket.yaml contains all the containers needed


Bairdo:
- A container's primary interface's MAC address can be set in the docker-compose using ```mac_address: 44:44:44:44:44:44```

anarkiwi:
Let's get just the Dockerfile in for the moment - please make a directory under docker, gasket, and put the Dockerfile.hostapd there.

Can you also get gcc, make hostapd and remove gcc in the same layer? This will make the image a lot smaller. And please if you can do that with alpine so it's even smaller again that'd be great.

anarkiwi:
Please add also a comment to why we need to use your fork of hostapd (summary of what features have been added and links to any issues you have open with them, if any).

Bairdo:
This version of hostapd is required to fix two bugs and adds one feature.
Bug 1 - identity returned with the control socket is null while using the internal radius server (not supported by feature 1). 
Bug 2 - in the same place sometimes the identity would have random characters off the end of the string (it's not null terminator-ed)
[help thread](http://lists.infradead.org/pipermail/hostap/2017-September/037933.html)
[patch thread](http://lists.infradead.org/pipermail/hostap/2017-September/037935.html) - fixes both bugs.


Feature 1 - save RADIUS attributes on Access-Accept
[thread](http://lists.infradead.org/pipermail/hostap/2017-September/037951.html)
A patch for this has not been submitted. And there is a possible memory leak of the attributes not being removed. iirc im not sure where the struct sta_info is destroyed.

Bairdo:
@anarkiwi updated dockerfile with alpine

gizmoguy:
Hi @Bairdo is it useful for me to start building this and the freeradius dockers on our docker build infrastructure so that users can pull these images down with docker without needing to build them manually?

Or shall I wait until you are a bit further through the project?

Bairdo:
hostapd b - https://github.com/Bairdo/gasket/issues/43

freeradius cmd update https://github.com/faucetsdn/faucet/pull/1878

Issue #1839, opened 2018-04-11
Use of Assert statements in Config verification


KitL:
Our configuration verification relies pretty heavily on assert statements. Assertions can be disabled at run time, which may not be an issue for us, but either way is the sort of bad coding practice that we shouldnt be using.

Instead we should just test things and in the case of failure raise our own exception.

anarkiwi:
https://github.com/faucetsdn/faucet/pull/1895 fixed.

PR #1895, opened 2018-05-02
Refactor abundance of assert statements

mab68:
Replace assert statements with a callable function that raises an InvalidConfigError if the condition parameter is true.
#1839


gizmoguy:
Hi @mab68 

Looks like this failed the tests due to decrease in pylint score for pylint ``faucet/config_parser.py``.

If you run pylint manually on this file it should tell you what you can do to improve the score.

PR #1918, opened 2018-05-09
added configuration documentation for meters

Bairdo:
From what i can tell this is correct, of course any suggestions and clarifications are welcome.


gizmoguy:
Hi @Bairdo 

Thanks for this!

RST has these nice ``note`` annotations you can use like:

.. note:: This is a note admonition.
   This is the second line of the first paragraph.

   - The note contains all indented body elements
     following.
   - It includes this bullet list.

Could you switch to using those instead of "Note:"

PR #2042, opened 2018-06-05
Implementation of probe protocol for failures on stack

trungdtbk:
The protocol will be able to detect wrong cabling, unidirection, and link broken. 
No handling of failures for the moment. 
It takes 2 intervals to detect if a link is working, 3 intervals (configurable) to confirm a link has broken.
The interval can be configured by lldp_beacon send_interval. However, the smallest value can be 5 seconds (the hardcoded interval for send_lldp_beacon service).
Next step to handle failures, including root bridge failure.

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -358,7 +358,7 @@ configuration.
         for dp in dps:
             if dp.stack is not None:
                 stack_dps.append(dp)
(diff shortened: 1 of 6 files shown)


anarkiwi:
Thanks for your contribution!

PR #2055, opened 2018-06-09
Prep methods for handling stack events

trungdtbk:
The purpose is to have Valves to update its stack topo upon network events (dp connect, disconect, stack port up/down). No update codes are implemented in this PR.
This prep is for making flood rule calculation dynamic to support root and other failures.


anarkiwi:
Can you explain please with an email to faucet-dev, what your proposed changes to the flooding and other algorithms might be?

PR #2125, opened 2018-07-01
Update stack graph according to the stack link probe

trungdtbk:
This does not change any of the stack flooding algorithm. It simply add/remove edges in the stack graph (networkx).

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -342,8 +342,9 @@ configuration.
         if port.lldp_beacon_enabled():
             self.lldp_beacon_ports.append(port)
 
-    def resolve_stack_topology(self, dps):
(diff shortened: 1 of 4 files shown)


anarkiwi:
This looks like a positive and welcome addition, but we need tests to be able to accept it.

trungdtbk:
Thanks for your comments. 
There's not much to test on this PR, I think. However, my next PR which implements the strategy as in email sent to Faucet-dev, will of course need tests.

anarkiwi:
We can't accept the PR sorry without tests, as this PR does change behavior when probes are enabled and may affect production deployments.

anarkiwi:
When tests are available, please update this PR (or open a new one).

PR #2135, opened 2018-07-05
Rename/reorganize test scripts

### LINKED ISSUE #2118: Refactor faucet test suite
This ticket will cover the first round of refactoring the test suite.

The goal at the end of this ticket is to have:

  1. Different tests (clib, unit tests, integration tests) groups together in a nice directory structure
  2. File names make it obvious what the file does and how to kick off each test

lantz:
Integration test scripts are moved into tests/integration/
Code checks are moved into tests/codecheck
Unit tests are moved into tests/unit

Removes obsolete tests in tests/config

Some other renaming for clarity:

clib_mininet_test_unit.py -> clib_mininet_tests.py
faucet_mininet_test_unit.py -> tests/integration/mininet_tests.py

clib/mininet_test.py -> clib/mininet_test_main.py

Note clib is a unit so the renamed files in clib/ change the API
slightly, possibly resulting in breakage for clients that import
these modules.

Fixes #2118


lantz:
Hmm this is odd - I am not sure why the CI tests would fail since they worked on my repo.

Also I'm not sure why the code coverage would have changed since as far as I can tell the same tests are running on the same code.

lantz:
Hmm, I'm not sure about the `coverage` issue! When I run it it comes out as 93% rather than 92.85%.

gizmoguy:
I'm fairly sure we have a coverage calculation bug as I see most PRs here seem to trip it by small percentage points (maybe we have a rounding issue somewhere).

I will now do a proper review of things here, make sure they work and get back to you.

lantz:
Oh I think I know what is wrong maybe? I think previously we were running the experimental API test app on its own as a unit test, but now we're not. Was this correct behavior?

PR #2136, opened 2018-07-05
Reset the hardware type to 'Open vSwitch' for the virtual switch

creneCSCO:
If we are running the test FaucetStackStringOfDPUntaggedTest with a hardware type that is sending
a table feature request (TFM) to start, the test will fail with an error because a table feature request 
will be sent as well to the virtual switch which could not handle this request.
The issue is coming from the fact that all the switches in the faucet yaml file will use the same hardware type instead of only the real switch should use the hardware type and the virtual switch should use the 
'Open vSwitch' type.


anarkiwi:
Thanks for your contribution! I just merged a patch from your email.

PR #2143, opened 2018-07-07
explicitly throw errors if they occur

cglewis:
this will ensure things like #2142 get caught.


cglewis:
Note, I expect this will fail until #2142 gets merged in and this branch is updated.

PR #2146, opened 2018-07-07
updates versions of influxdb and grafana

cglewis:
brings influxdb up to 1.6 for both amd64 and arm compose files.  opted away from hypriot's image as it's now archived and hasn't been updated in a year (https://github.com/hypriot/rpi-influxdb).

brings grafana up to the latest for amd64, no newer version for arm yet.


gizmoguy:
I was looking into the grafana case for arm the other week and (fg2it/grafana)[https://github.com/fg2it/grafana-on-raspberry] is now end of life stating grafana officially do arm builds themselves now, however I haven't seen an official docker arm image pop out yet.

cglewis:
Huh, I haven't seen one yet either.

PR #2180, opened 2018-07-19
configuration inconsistency fix #1836

### LINKED ISSUE #1836: Configuration inconsistency between lldp org_tlvs and routes
When we have a list of configuration elements sometimes we list the configuration blocks directly and sometimes we keep them in a second dictionary.

This is needlessly unintuitive.

For example when we do things directly it looks like:
```
                    org_tlvs:
                        - {oui: 0x12bb, subtype: 2, info: "01406500"}
```

Or when we keep things in an additional dictionary it looks like this:
```
        routes:
            - route:
                ip_dst: 10.99.99.0/24
                ip_gw: 10.0.0.1
```

To me the org_tlvs approach seems more sensible. I propose we migrate to that approach (by first allowing both options).

mab68:
allows vlan routes to be configured like org_tlvs as suggested in #1836:
```
        routes:
            - {ip_gw: '10.0.0.1', ip_dst: '10.99.99.0/24'}
```

### DIFF
--- a/faucet/vlan.py
+++ b/faucet/vlan.py
@@ -253,16 +253,19 @@ class VLAN(Conf):
                 'Must be as many BGP neighbor addresses as BGP server addresses'))
 
         if self.routes:
+            test_config_condition(not isinstance(self.routes, list), 'invalid VLAN routes format')
             try:
                 self.routes = [route['route'] for route in self.routes]
             except TypeError:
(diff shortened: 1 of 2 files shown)


gizmoguy:
@mab68 could you please add an additional test case to the test suite to test the new format of routes work?

mab68:
@gizmoguy added additional test cases

PR #2198, opened 2018-07-23
Update stack-graph according to linkstate events

trungdtbk:
This is to reflect link-state up/down events on a stack into each DP's stack graph. It works when all DPs are controlled by the same Faucet or events are local (no mechanism to distribute events across Faucet yet).
The updated graph is then used to re-calculate flood/unicast rules.


anarkiwi:
why is "has_changed" necessary?

trungdtbk:
It is used because some valve services don't affect the topo and they don't change to topo for every execution.

anarkiwi:
You could call the affected valve's send_msg immediately and not create another state variable - it will also converge faster.

anarkiwi:
If you have eliminated the has_changed variable, we can merge this.

trungdtbk:
Yes. I will update that. I'm building a mininet to test controller restart working properly.

anarkiwi:
Yes, it can have access to other_values in the same way as packet et al does.

PR #2257, opened 2018-08-08
Add faucet Raspberry Pi image documentation/tutorial

mab68:


gizmoguy:
I see this is merged already but I have reviewed and there are some changes required in an additional PR:

  * "facuet" is spelt wrong
  * Some of the RST is broken (I see ``:caption`` and ``:name`` haven't worked in some place)
  * I see a lot of duplication between this tutorial and the "Installing faucet for the first time tutorial". I think we should go with my original suggestion of not having an additional tutorial for this but adding a "Installing on raspberry pi" secontion to the Installation chapter and linking into the right subsections of the "Installing faucet for the first time tutorial".

PR #2294, opened 2018-08-15
Acl port resolution

KitL:
Fix for https://github.com/faucetsdn/faucet/issues/2058

Also tidies constructor for ACL object to make it more consistent with other conf objects (uses conf __init__ to set valuables and check_config for validation rather than doing it in __init__)

### DIFF
--- a/faucet/acl.py
+++ b/faucet/acl.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 import copy
 
+from faucet import valve_of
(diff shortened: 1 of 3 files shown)


KitL:
It occurred to me that the ability to configure acls by omitting the 'rule' keyword to be consistent with https://github.com/faucetsdn/faucet/issues/1836 didnt exist before these commits, so I added a test for that as well.

PR #2303, opened 2018-08-18
Update mininet-based integration tests to python 3

lantz:
There are a lot of changes here so you will probably want to review them - I can definitely help with any clarification or additional requested changes.

### Notes

These changes migrate FAUCET's Mininet-based integration tests to Python 3, removing the requirement for Python 2.

They depend on the test base docker image changes in faucetsdn/docker-test-base#4.

**Overview**

- We install a Python 3-compatible version of Mininet in the base test image (also see below)
- We remove the Python 2 pip dependencies and install the Python 3 versions
- The Mininet-based integration tests are migrated to Python 3
- We invoke the integration tests using Python 2 rather than Python 3
- We run `pytype` invoking it with `-V3.5` on all code
- We invoke a workaround in `docker/workarounds.sh` to compensate for an upstream `ptype` issue (see below)

**Potential Issues**

- `FaucetTcpdumpHelperTest-test_tcpdump_nextline: not all controllers healthy` on shard 3 (see below)
- This patch currently skips the IPv6 `test_fuzz_controller()` test which seems to hang consistently on travis but not locally; it would be nice to find and fix the root cause of this behavior
- Python 3 `pytype` chokes on `testtools` with a fatal error that we cannot catch, so this patch breaks `pytype.sh` unless the fix in `workarounds.sh` is applied. See google/pytype#133
- This may slow down some CI tests somewhat (`sanity` shard seems to be several minutes slower in my testing); we may wish to investigate this to examine the root cause and see whether it can be improved

**Subsequent Work**

- After this is merged into mainline, we can remove Python 2 support from the base test image
- We might want to install Mininet in `run_tests.sh` instead of the base image
- I still would like to remove the Mininet hacks/workarounds in `mininet_topo.py` at some point, as they should no longer be necessary (and should also be fixed in Mininet mainline if they aren't already!)

### DIFF
--- a/clib/clib_mininet_test.py
+++ b/clib/clib_mininet_test.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 """Mininet tests for clib client library functionality.
 
--- a/clib/clib_mininet_test_main.py
+++ b/clib/clib_mininet_test_main.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 """Mininet test runner
 
@@ -28,7 +28,7 @@ import subprocess
 import tempfile
 import threading
 import time
-import unittest2
+import unittest
 
 import yaml
 
@@ -117,7 +117,7 @@ def import_hw_config():
         required_config = {
             'dp_ports': (dict,),
             'cpn_intf': (str,),
-            'dpid': (long, int), # pytype: disable=name-error
+            'dpid': (int),
             'of_port': (int,),
             'gauge_of_port': (int,),
         }
@@ -163,9 +163,9 @@ def check_dependencies():
                 stderr=subprocess.STDOUT,
                 close_fds=True)
             proc_out, proc_err = proc.communicate()
-            binary_output = proc_out
+            binary_output = proc_out.decode()
             if proc_err is not None:
-                binary_output += proc_err
+                binary_output += proc_err.decode()
         except subprocess.CalledProcessError:
             # Might have run successfully, need to parse output
             pass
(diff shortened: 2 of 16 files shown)


lantz:
@anarkiwi @gizmoguy So I've opened a pull request for this so we can start reviewing it and working out any remaining issues. It won't work yet because it depends on my PR to the base image, but once that is merged and builds on dockerhub it should hopefully pass travis ci testing. Let me know if you have any questions or run into any issues!

tests/codecheck/src_files.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 FAUCETHOME=`dirname $0`"/../.."
-for i in clib faucet tests ; do find $FAUCETHOME/$i/ -type f -name [a-z]*.py ; done | xargs realpath | sort
+for i in clib faucet tests ; do find $FAUCETHOME/$i/ -type f -name '[a-z]*.py' ; done | xargs realpath | sort
```
lantz: This isn't strictly necessary for this patch, but it seems like a good idea. I may open a separate PR for it.

lantz:
Hmm, surprising that the tests failed. They worked for me:
https://travis-ci.org/lantz/faucet/builds/418090070

lantz:
Hmm, seeing 
    FaucetTcpdumpHelperTest-test_tcpdump_nextline: not all controllers healthy
but it doesn't break the ci tests, oddly enough. It doesn't seem to happen locally for me though?

lantz:
OK, with the `DockerHost` fix it seems to pass!

anarkiwi:
Since we've extensively discussed this OOB - let's get the change in, and mopping up can be done separately. You have done the vast majority of the infrastructure change - fantastic! dotting i's can be done next.

lantz:
@anarkiwi Wow, great!  Thanks. ;)

There are a few things I want to look at (including some issues above) and also let me know if there are any other issues you run into with these changes.

Ah, thanks for running/fixing the HW tests! I didn't have an infrastructure for testing that, though it occurs to me that I could think about doing that if I wanted to - perhaps using public infrastructure like GENI and/or setting up a mini test lab of my own.

PR #2316, opened 2018-08-21
Handle -c flag properly

lantz:
It seems that the `-c` flag/`$CODECHECK` setting wasn't being handled properly.


lantz:
oops this is wrong - brain not working I think!

PR #2323, opened 2018-08-22
Chewie integration tests

Bairdo:
I've created this new dict variable (HOST_NAMESPACE) to be used to set if a host should be in the root namespace or not. At the moment it only works with untagged hosts. But i can extend it to work with the others if that's wanted?

An eth_dst rewrite (to nfv_host direction) to any MAC also works (this PR is just using the reserved address).

### DIFF
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -81,6 +81,7 @@ class FaucetTest(mininet_test_base.FaucetTestBase):
 class FaucetUntaggedTest(FaucetTest):
     """Basic untagged VLAN test."""
 
+    HOST_NAMESPACE = {}
     N_UNTAGGED = 4
     N_TAGGED = 0
     LINKS_PER_HOST = 1
@@ -111,7 +112,8 @@ vlans:
         self.topo = self.topo_class(
             self.OVS_TYPE, self.ports_sock, self._test_name(), [self.dpid],
             n_tagged=self.N_TAGGED, n_untagged=self.N_UNTAGGED,
-            links_per_host=self.LINKS_PER_HOST, hw_dpid=self.hw_dpid)
+            links_per_host=self.LINKS_PER_HOST, hw_dpid=self.hw_dpid,
+            host_namespace=self.HOST_NAMESPACE)
         self.start_net()
 
     def verify_events_log(self, event_log):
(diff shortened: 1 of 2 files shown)


tests/integration/mininet_tests.py
```diff
@@ -165,20 +167,20 @@ class FaucetUntagged8021XTest(FaucetUntaggedTest):
...
                 output:
-                    set_fields:
-                        - eth_dst: 00:00:00:00:00:01
+                    # set_fields:
+                        # - eth_dst: NFV_MAC
                     port: b4
         - rule:
             actions:
-                allow: 1
+                allow: 0
```
Bairdo: not sure why i changed this. both true or false work

anarkiwi:
Good progress here.

Hard coded sleep values will make the tests brittle under load. Can you add some other checks to make sure the process has started? tcpdump_helper is one way to do this.

Bairdo:
Ok. I've found that the EAP Success message is not actually getting sent, and that there is an exception in the auth_handler callback, which does not get caught/logged.

Do you know why this would be occurring?

```
...
Aug 23 23:11:45 faucet INFO     Successful auth from MAC 86:31:68:c0:33:33 on Port 1
Aug 23 23:11:45 faucet ERROR    eth_src match in table port_acl
  File "/faucet-src/faucet/faucet_dot1x.py", line 61, in auth_handler
    self.dot1x_port.number, str(address))
  File "/faucet-src/faucet/valve.py", line 1446, in add_authed_mac
    inst=[valve_of.goto_table(self.dp.tables['vlan'])])
  File "/faucet-src/faucet/valve_table.py", line 105, in flowmod
    self._verify_flowmod(flowmod)
  File "/faucet-src/faucet/valve_table.py", line 67, in _verify_flowmod
    '%s match in table %s' % (match_type, self.name))
AssertionError: eth_src match in table port_acl
```
fyi:
self.match_types = {'in_port': False, 'eth_type': False}

anarkiwi:
OK good - tests find problems! This is because the auth handler has had no test coverage/usage. It's trying to add a flow with an eth_src match, but the pipeline verifier says that it doesn't know that the port_acl table is provisioned for that match (this is part of the code infrastructure that tells TFM switches what they need).

For now, make sure you have port ACLs applied always, that always match eth_src. We can evaluate how better to handle this (e.g. the pipeline provisoner can always add eth_src if dot1x is enabled).

Bairdo:
see #2357 for Prometheus variables.

anarkiwi:
You've helped us make another step forward, so I merged and worked around
the minor lint problem for the moment.

Would be great to see port level variables too!

PR #2347, opened 2018-08-24
Add tests that config options are generated correctly

KitL:
Slowly working my way through the config tests


KitL:
Involves moving all the correct config tests to the top of the file, which is why so many changes.

PR #2357, opened 2018-08-26
Chewie prometheus metrics

Bairdo:
Adds a prometheus success/failure counter.
failure isn't currently used, will require a change to chewie - perhaps a failure_handler


Bairdo:
these counters are at the dp level. would at the port level be better?

I'm thinking we also want to know who (mac/user) is currently authenticated, but not sure we can do that with Prometheus.

What other metrics are useful? 
- number of active clients?

anarkiwi:
Port level would be better, yes please.

I see the build failed due to declining lint score in dot1x - please if you would look at that too.

PR #2377, opened 2018-08-28
dot1x tests with chewie v0.0.5 (radius & state handlers).

Bairdo:
Tests check all dot1x prometheus variables.

Tests must run serial as cannot configure chewie's radius port

PR #2380, opened 2018-08-30
Chewie multi valves

Bairdo:
Multiple ports should be able to use 802.1x.
The first Valve's config decides the nfv interface chewie will listen on.

Builds upon #2377, and goes with https://github.com/faucetsdn/chewie/pull/22


anarkiwi:
Looks like merge conflicts to clean up here.

PR #2381, opened 2018-08-30
Next table properties - *please check next tables are correct*

KitL:
Tidies up generation of table objects and uses correct next table properties in TFMs rather than just listing every subsequent table.

I have added tests that the Conf objects are generated with the correct next tables. And the code works against an Aruba. But it seems entirely possible that that is because they ignore the next tables property.

So if you could just double check the test cases that my expected next tables are correct.
These are the two tests, one with a full pipeline and one with a minimal pipeline.
https://github.com/faucetsdn/faucet/compare/master...KitL:next_table_properties?expand=1#diff-f0b4874025da0a363291e3ea22782aedR881
https://github.com/faucetsdn/faucet/compare/master...KitL:next_table_properties?expand=1#diff-f0b4874025da0a363291e3ea22782aedR1035


anarkiwi:
Looks like a minor lint problem?

anarkiwi:
Thanks! Aruba do seem to check the next table property - when I got it wrong in the past it would reject flows.

PR #2385, opened 2018-08-31
Actually use the next tables - *probably needs to be run against hardware again*

KitL:
Turns out after calculating the next tables properly, I never actually changed the message that gets sent. So the testing we did needs to be repeated.


KitL:
passes against the aruba

PR #2412, opened 2018-09-03
Dot1x config

Bairdo:
requires https://github.com/faucetsdn/chewie/pull/25

chewie config options for radius server ip/port/secret & chewie id


tests/integration/mininet_tests.py
```diff
@@ -272,6 +275,8 @@ class FaucetSingle8021XSuccessTest(FaucetUntaggedTest):
 
     HOST_NAMESPACE = {3: False}
 
+    RADIUS_PORT = 1840
```
Bairdo: port(s) should ideally be random free ports.

anarkiwi:
Looks like some merge conflicts here.

PR #2419, opened 2018-09-04
Revert revert chewie

Bairdo:


anarkiwi:
Looks like pylint is failing - because primarily, catching Exception is too risky.

PR #2449, opened 2018-09-13
Static eth src

KitL:
Progress towards making the eth_src table static

creates a classification table before eth_src for directing packets to FIBs, filtering bogons and blocking mac spoofing.

The eth_src table still handles flows for override_output_port (see https://github.com/faucetsdn/faucet/issues/2448) and has vlan matches for the controller flows.


anarkiwi:
This a really nice change but we can't use it yet based on the limited table space for SCINet. 
Could we make this incremental somehow (e.g. have an alias for classification table that points to eth_src in the meantime)?

Issue #2448, opened 2018-09-13
override_output_port


KitL:
I am looking at tidying up the eth_src table.

The override_output_port option is a bit of a nuisance for this. It was created in response to https://github.com/faucetsdn/faucet/issues/1754
But it only affects layer 2 packets at the moment, which seems like not what was asked for. Since no one has noticed that I am wondering how much anyone is actually using this function?
@cglewis @Bairdo you guys said you were interested in this.

To make this work properly, we need a second extra table (I am already adding a table to move most of the eth_src stuff into). Or we can wait until we add egress ACLs (I actually do plan on adding these next or next-ish) when this kind of thing becomes a bit less haxy.

So how important is this functionality? Is it worth adding an extra table for?

cglewis:
I admittedly haven't used this feature very often, but it was handy the few times I have.  So I'm not sure it's necessarily important enough for an extra table.

anarkiwi:
There is a generic L2 tunnel facility available now (even works between endpoints on the same DP).

Shall we remove this old override feature in favor of tunnelling?

PR #2477, opened 2018-09-18
Static eth src option

KitL:
Progress towards making the eth_src table static

creates an optional classification table before eth_src for directing packets to FIBs, filtering bogons and blocking mac spoofing.

The eth_src table still handles flows for override_output_port (see #2448) and has vlan matches for the controller flows.

This can be configured with the key 'use_classification' in the dp configuration.

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -187,7 +187,7 @@ configuration.
         'port_acl': int,
         'vlan': int,
         'vlan_acl': int,
-        'use_classification': int,
+        'classification': int,
         'eth_src': int,
         'ipv4_fib': int,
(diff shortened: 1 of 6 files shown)


anarkiwi:
default_table_sizes_types = {
￼         'port_acl': int,
￼         'vlan': int,
￼         'vlan_acl': int,
￼         'use_classification': int,
￼         'eth_src': int,
￼         'ipv4_fib': int,
￼         'ipv6_fib': int,

I think you mean 'classification' here.

KitL:
This passes locally. Is this a random flake by any chance?

PR #2526, opened 2018-09-27
Valve pipeline 1

KitL:
Include the valve_pipeline module to handle the movement of packets through the pipeline and update valve_route to remove all references to the classification_table by using the new module in its place.


anarkiwi:
Obsoleted by https://github.com/faucetsdn/faucet/pull/2537.

PR #2527, opened 2018-09-27
Valve host pipeline

KitL:
Builds on https://github.com/faucetsdn/faucet/pull/2526

Removes references to classification table from valve host by using the pipeline module.

This involves a change to the functionality of permanent learn.

Permanent learn now only blocks spoofed mac addresses when they are seen by the controller. Rules blocking the port/mac combination are installed in response to receiving any packet from a permanent learn mac address on a different port.

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -81,6 +81,7 @@ class Valve:
         'dp',
         'flood_manager',
         'host_manager',
+        'pipeline',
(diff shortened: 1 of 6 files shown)


anarkiwi:
Obsoleted by https://github.com/faucetsdn/faucet/pull/2538.

PR #2577, opened 2018-10-12
added 802.1x documentation

Bairdo:
Does anyone think it should be explicitly stated how to connect things together? maybe with a diagram?

e.g.

```
sw.nfv_sw_port ------ nfv_intf.Faucet
sw.controlplane === Faucet

```


Bairdo:
Hi @gizmoguy 

I've added a basic diagram, but without the radius side. How should Faucet connect to the RADIUS server, via separate connection or via the switch's dataplane?

gizmoguy:
I expect the RADIUS server will be available to the controller itself on a separate NFV port via the normal linux routing table. This NFV interface of course could be served via a faucet or non-faucet network. For smaller deployments we could also expect RADIUS to run on the same machine as faucet.

PR #2579, opened 2018-10-12
802.1x actually block unauthenticated traffic

Bairdo:
I'd screwed up the ping fail check. Sorry. :(

Will try add a tcpdump check, before i leave shortly or monday.

https://travis-ci.org/Bairdo/faucet/builds/440465340


Bairdo:
#2584 removes the assertion catching. Turns out one_ipv4_ping() can already fail if ping is not successful.

PR #2583, opened 2018-10-15
Fix raspbian builds.

gizmoguy:
* Use latest release of pi-gen rather than master.
  * Stage5 no longer exists.


gizmoguy:
This runs okay on my branch: https://gitlab.com/gizmoguy/faucet/-/jobs/107962895

PR #2588, opened 2018-10-16
Correct handling when 802.1X nfv port changes state

Bairdo:
previously if the nfv port went down and up, the flows to forward between supplicant and nfv would not be added to the nfv port.


anarkiwi:
Looks like this is failing python3.4 unit tests?

PR #2602, opened 2018-10-29
Run all tests on Travis on new branch.

anarkiwi:


gizmoguy:
We should also run the docs test when files in ``docs/`` changes.

They can be run like this:

```
  cd ./docs
  make html || exit 1
  rm -rf _build
```

We will also want to skip running the docs test inside the docker tests.

PR #2613, opened 2018-11-02
8021X hardware tests

Bairdo:
Running through full integration tests on cisco c9300 now. - b19cc8a
Will try to test it on a AT x510 as well, but waiting on some NICs/serial cable.

The problem seems to be that ACL.matches and ACL.set_fields dicts do not get populated until the ACL is 'built' (ACL.build()). However in https://github.com/faucetsdn/faucet/blob/eba318918bbfa75809ed2977b68dc08d9449e213/faucet/dp.py#L363 a 'dummy' ACL is added (PORT_ACL_8021X) to the generate acl tables, but the ACL is never built.

~~I will try simplify this solution out to either just build the dummy ACL before it gets used, or build all ACLs in init~~ done in https://github.com/faucetsdn/faucet/pull/2613/commits/fb95f7d64d2d6ea13c339e1042c8c6706025961c

### DIFF
--- a/faucet/acl.py
+++ b/faucet/acl.py
@@ -117,6 +117,18 @@ The output action contains a dictionary with the following elements:
             conf['rules'].append(rule.get('rule', rule))
         super(ACL, self).__init__(_id, dp_id, conf)
 
+        for rule_conf in self.rules:
+            self.matches['in_port'] = True
+            for k in rule_conf.keys():
+                if k != 'actions':
+                    self.matches[k] = True
+            if 'actions' in rule_conf:
+                if 'output' in rule_conf['actions']:
+                    if 'set_fields' in rule_conf['actions']['output']:
+                        _set_fields = rule_conf['actions']['output']['set_fields']
+                        for field_pair in _set_fields:
+                            self.set_fields.add(list(field_pair.keys())[0])
+
     def check_config(self):
         test_config_condition(
             not self.rules, 'no rules found for ACL %s' % self._id)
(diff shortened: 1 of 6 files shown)


Bairdo:
I don't think building all ACLs would be correct as we don't know if the ACL is on a port or VLAN. And if building all ACLs in the __init__ we would need to give it a port (so the 1x case is covered) so the 'in_port' match is return for table generation which is not necessarily correct when only VLAN acls are used. - this does however pass the software tests.

So fb95f7d only builds the 8021X acl.

Bairdo:
https://travis-ci.org/Bairdo/faucet/builds/449627071

Bairdo:
fb95f7d - running on the AT x510 the 8021X tests pass except for Faucet8021xPortChangesTest. FaucetSanityTest.test_untagged (also on master) also fails when flapping ports, so thinking something isn't right with the switch configuration. Haven't run through whole test suite.

anarkiwi:
Re-testing with latest switch firmware and Chewie 0.0.10.

gizmoguy:
@MichaelWasher ran the Faucet8021XConfigReloadTest-test_untagged, Faucet8021XFailureTest-test_untagged, Faucet8021XPortChangesTest-test_untagged and Faucet8021XSuccessTest-test_untagged on the following hardware and it worked fine:

cisco-9300 (firmware version 16.10.20180930:011424)
aruba-2930f (firmware version WC.16.05.0000GO5)
at-x930 (firmware version 5.4.8)

PR #2645, opened 2018-11-29
Use port.name for all port= labels and use common code between FAUCET and Gauge.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2640

PR #2663, opened 2018-12-05
Hash config files for changes if stat didn't change.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2649

PR #2668, opened 2018-12-07
Remove ZodiacFX tests, which were never really used (and Zodiac switc…

anarkiwi:
…hes now not commercially supported).


gizmoguy:
I believe there is some more we can cut out too, in particular:

  * ``tests/run_zodiac_integration_tests.sh`` can be deleted
  * ``hw_switch_config.yaml`` mentions of FaucetZodiac can be removed

PR #2677, opened 2018-12-10
remove pointless delete messages

KitL:
We never set the out_port on any flow table entry, so deleting things that match that out_port is a no op.


KitL:
I'll actually add this to another PR to fix a conflict

PR #2684, opened 2018-12-11
Re-enable groups for stacking.

anarkiwi:
Don't send group with same actions but different ID for unicast flooding (may cause hash collision).
Add more port flap tests to stacking.


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2682

PR #2695, opened 2018-12-12
Include unit test to catch faulty instructions

KitL:
check flow mods dont have duplicate instructions


gizmoguy:
Rebase on faucet@master to get your tests passing again.

PR #2703, opened 2018-12-18
WIP: Add Dynamic ACLS for 802.1X Ports

MichaelWasher:
Add Default (Failure) ACL Functionality
Add Login (Success) ACL Functionality
Port Down Remove ACL

PR #2697, opened 2018-12-12
BUGFIX: Chewie port never comes back up

MichaelWasher:
Add call to Chewie port_up

PR #2722, opened 2019-01-10
acl manager

KitL:
I've created an acl manager to control the installation and removal of flows from the acl tables.

This affects ACLs and dot1x obviously.

The only functional changes are a change to the priorities used in the acl table, so that they are hard coded rather than configurable. Also the no ACL rule doesnt get added to output only ports now, so they will get dropped in the ACL table rather than the vlan table (unless other ACLs are added to that port).

@Bairdo can you have a look over the logic here and ensure that what I am doing makes sense. I guess the main thing to notice is that ACLs and the flows installed when a dot1x port comes up are the same priority value.


gizmoguy:
Will this break #2703 and #2720 ?

faucet/faucet_dot1x.py
```diff
@@ -144,124 +141,87 @@ def set_mac_str(self, valve, valve_index, port_num):
...
+            nfv_sw_port (Port):
+            acl_manager (ValveAclManager):
 
-    def create_flow_pair(self, dot1x_port, nfv_sw_port, valve):
-        """Creates the pair of flows that redirects the eapol packets to/from the supplicant and
-        nfv port
+        Returns:
+            list of flowmods
+        """
+        self.dot1x_speaker.port_down(
```
Bairdo: This was an error on my part and should have been port_up, think @MichaelWasher has fixed it in pr #2697
But seeing how this PR is not adding functionality, I'm guessing someone will refactor later.

faucet/faucet_dot1x.py
```diff
@@ -144,124 +141,87 @@ def set_mac_str(self, valve, valve_index, port_num):
...
+            nfv_sw_port (Port):
+            acl_manager (ValveAclManager):
 
         Returns:
             list of flowmods
         """
-        self.dot1x_speaker.port_down(get_mac_str(self.dp_id_to_valve_index[valve.dp.dp_id],
-                                                 dot1x_port.number))
-        nfv_sw_port = valve.dp.dot1x['nfv_sw_port']
+        self.dot1x_speaker.port_down(
```
Bairdo: (This (port_down()) was an error on my part and should have been port_up(), think @MichaelWasher has fixed it in pr #2697
But seeing how this PR is not adding functionality, I'm guessing someone will refactor later.)

Thinking about when the NFV port goes up.
The system is effectively down, we miss logoff/new authentication attempts coming from the supplicants but should still get port events via OpenFlow.

i think there are a few options we could do:

A) Don't notify Chewie as it does not currently do anything if the NFV port is down.
B) Notify Chewie, and later we can do something (e.g. Send a preemptive Identity Request). We could interpret the NFV port being down as meaning that everyone is deauthenticated, but I'm not sure if we would want to do this.

I think option A is fine for the moment - so this line should be able to be removed, and it shouldn't affect functionality.

KitL:
It doesnt look like it breaks Mark's PR, but it will definitely break Michael's one. Though Michael's one already needs to be refactored it seems.

I'll go through Michael's work and look at what needs to be changed.

Also I dont intend to write unit tests for the dot1x stuff. It would be nice if there were valve tests for dot1x added though. It did lead to a bit of frustration for me.

gizmoguy:
Can we discuss on the team meeting call today about what to do with this PR - since we need to come up with a good strategy for PR #2703

KitL:
I'll fix the conflicts in 2703

PR #2733, opened 2019-01-18
update pika version; use alpine image for management

cglewis:


anarkiwi:
github seems confused. travis had succeeded but the status check is stuck on queued.

no obvious way to unstick it. would you close this PR and reopen with the same change? sorry about that.

Issue #2759, opened 2019-01-28
Faucet IVR breaks when hosts move port and faucet is sent SIGHUP


gizmoguy:
Let's say we have host1 (10.0.0.1/24) and host2 (10.0.1.1/24).

Let's boot a clean faucet with the following configuration:

```
vlans:
    vlan100:
        vid: 100
        faucet_vips: ["10.0.0.254/24"]
        faucet_mac: "00:00:00:00:00:11"
    vlan200:
        vid: 200
        faucet_vips: ["10.0.1.254/24"]
        faucet_mac: "00:00:00:00:00:22"
routers:
    router-1: 
        vlans: [vlan100, vlan200]
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        interfaces:
            1:
                name: "host1"
                description: "host1 network namespace"
                native_vlan: vlan100
            2:
                name: "host2"
                description: "host2 network namespace"
                native_vlan: vlan200
```

IVR works fine and all hosts can ping. Now if we move host2 from port 2 to port 3, update the configuration to reflect this and reload faucet we will find the route to 10.0.1.1/32 is still pointing to port 2 which no longer exists and we have to wait until this timeouts for this host to rejoin the network.

gizmoguy:
This still seems to be a problem after #2762.

When I am testing the host move from port 2 to port 3 I am destroying and recreating the whole bridge, wondering if this is related to faucet missing the port down event.

I have captured two sets of logs, one in the case where it works (we explicitly take port 2 down before deleting the bridge):

> Jan 30 10:04:45 faucet INFO     Reloading configuration
> Jan 30 10:04:45 faucet INFO     configuration /etc/faucet/faucet.yaml changed, analyzing differences
> Jan 30 10:04:45 faucet INFO     Add new datapath DPID 1 (0x1)
> Jan 30 10:04:45 faucet.valve INFO     DPID 1 (0x1) sw1 IPv4 routing is active on VLAN vlan100 vid:100 untagged: Port 1 with VIPs frozenset({IPv4Interface('10.0.0.254/24')})
> Jan 30 10:04:45 faucet.valve INFO     DPID 1 (0x1) sw1 IPv4 routing is active on VLAN vlan200 vid:200 untagged: Port 2 with VIPs frozenset({IPv4Interface('10.0.1.254/24')})
[...]

PR #2765, opened 2019-01-30
Force switch OF reconnection on a cold start. This ensures that FAUCET and the switch have the same list of ports.

anarkiwi:
Previously we could be desynchronized (e.g. adding a port, but we don't know its oper status).
Add internal API change to allow a Valve to request a switch OF reconnection.


gizmoguy:
#2759

PR #2766, opened 2019-01-31
Tunnel tests don't run on hardware due to hard coded port references.

anarkiwi:


gizmoguy:
Looks like this PR should have a companion github issue attached that will track fixing this issue properly.

anarkiwi:
On re-test is fixed by previous weeks' PRs.

PR #2784, opened 2019-02-07
Implement error signal in Prometheus for invalid config.

anarkiwi:


gizmoguy:
I think the proposed "faucet_config_valid" could be confusing as to which config file it is talking about. Faucet will always keep a running configuration in memory that works no matter what the on disk configuration file looks like.

How about a "faucet_configuration_load_error" option which triggers when faucet failed to correctly parse the configuration file it was just told to load.

PR #2793, opened 2019-02-13
Fix: 'Bugfix: Chewie port never comes back up'

Bairdo:
update of PR #2697 works on software but not hardware, see comment https://github.com/faucetsdn/faucet/pull/2697#issuecomment-462489499 for failure messages.


Bairdo:
All 1X tests pass on software, x510 & cisco 9300

PR #2802, opened 2019-02-24
docs/architecture: remove table numbers

ewenmcneill:
As suggested by https://github.com/faucetsdn/faucet/issues/2797#issuecomment-464518340 remove the "Table NN: " prefixes from each of the table headings (replacing them with "FOO Table") to make the names more prominent than the Table numbers used.

The diagram referenced from `architecture.rst` is also clearly out of date (eg, https://github.com/faucetsdn/faucet/issues/2797#issuecomment-463473745), but it appears to have been generated by some external tool with no source file available, so it is not updated.   Possibly someone with the source file for that diagram could update it, and remove the table ID numbers?

The `faucet-pipeline.png` diagram is at:

https://github.com/faucetsdn/faucet/blob/master/docs/_static/images/faucet-pipeline.png

and there's a `faucet-pipeline.txt` in the same directory:

https://github.com/faucetsdn/faucet/blob/master/docs/_static/images/faucet-pipeline.txt

Both were updated by the same commit, but it's unclear to me (a) how `faucet-pipeline.png` was generated, and (b) if `faucet-pipeline.png` was generated from the `faucet-pipeline.txt` file. 

At a glance I think `faucet-pipeline.txt` is ASCII art intended to represent the same thing as `faucet-pipeline.png`, both of which are now out of date, but there doesn't seem to be generation of one from the other.  `faucet-pipeline.txt` also seems to be unreferenced by the rest of the documentation:

```
(venv) ewen@faucet:~/faucet/docs$ find . -type f | xargs grep faucet-pipeline.txt
(venv) ewen@faucet:~/faucet/docs$ 
```

in which case possibly `faucet-pipeline.txt` should be removed?

### DIFF
--- a/docs/architecture.rst
+++ b/docs/architecture.rst
@@ -22,23 +22,27 @@ See unit and integration tests for working configuration examples.
 Faucet Openflow Switch Pipeline
 -------------------------------
 
-This summarizes the global FAUCET pipeline; however, certain tables may be omitted (adjusting table IDs accordingly)
+This summarizes the global FAUCET pipeline; however, certain tables may be omitted
 if the functionality is not required. For example, if routing is not configured, neither FIB table nor the VIP table
 will be provisioned.
 
-See also canonical pipeline definitions in faucet_pipeline.py
+Usually the OpenFlow table IDs will be allocated sequentially for the
+tables actually used, so tables should be referenced by their name
+rather than the table ID in this diagram.
+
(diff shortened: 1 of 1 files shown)


byllyfish:
I will take care of removing the table numbers from the faucet-pipeline.png diagram. It's just a hand-drawn version of the ASCII art version.

PR #2806, opened 2019-02-24
Remove table numbers from pipeline diagram.

byllyfish:
Per https://github.com/faucetsdn/faucet/pull/2802#issue-255657306 from @ewenmcneill:

Source is a google doc:
https://docs.google.com/drawings/d/1YbADYGfwx6fVHlnHh1YoabmojjtibI1dw4y77iuFRcI/edit?usp=sharing

This is still the same diagram from 2017 but with table numbers removed. That diagram was based on faucet-pipeline.txt ASCII art.

A diagram generated from the actual dynamic pipeline is a better future direction than this.

This is still missing some things:
1.  the vlan table punts lacp/lldp to the controller if those features are enabled.
2. this doesn't show the transitions out of port_acl that skip the vlan table. (I think that's only possible with force_port_vlan)
3. Other transitions from faucet_pipeline are not shown; I think a lot of these require certain features to be enabled.

An auto-generated diagram that shows the actual transitions (dependent on features enabled) seems to be the way to go.


ewenmcneill:
@byllyfish Thanks!  That looks consistent with the changes to the text in https://github.com/faucetsdn/faucet/pull/2802.

As a thought, would it make sense to export the SVG version out of the Google Docs diagram and include that in the Faucet repo too?  That'd at least be a potentially editable form in some other tool.   (And/or include the URL of the Google Docs, for ease of future reference.  Although it looks like there's no trivial way to "fork" the document in Google Docs, and have someone else edit it, but at least knowing where it came from could be useful later.)

I do agree that if we can get a diagram generation tool to automatically generate a reasonably laid out  diagram, as discussed in the thread starting at https://github.com/faucetsdn/faucet/issues/2797#issuecomment-463536734, then that'd be the best long term option.

Ewen

PR #2837, opened 2019-03-06
Reduce integration test run time for stacking tests by making LLDP block check more efficient

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2800.

Issue #2372, opened 2018-08-28
Sunset python2 support


KitL:
eg. https://github.com/faucetsdn/faucet/blob/1fd928a7eed637a3109ab591df460181e1cdf283/faucet/acl.py#L127

We convert this into a list unnecessarily so that the types are consistent between python 2 and 3. Now that we have deprecated python 2 we should not do this sort of thing.

lantz:
As I understand it, Python 2's `iteritems()` provided an iterator over a snapshot of the collection, while Python 3's `items()` iterator doesn't, so using `list(x.items())` is the conservative replacement for `iteritems()` and that is what `2to3` uses. Usually it isn't necessary, however, since most loops over a collection don't add or delete items in the loop body.

gizmoguy:
Reopening because we haven't added ``python_requires='>=3.4'`` to setup.cfg or setup.py yet.

gizmoguy:
pip versions < 9.0 ignore python_requires so we should explicitly check the python version inside faucet and abort when running on python 2

```
import sys

if sys.version_info < (3,):
    raise ImportError(
    """You are running Frobulator 6.0 on Python 2

Frobulator 6.0 and above are no longer compatible with Python 2, and you still 
ended up with this version installed. That's unfortunate; sorry about that. 
It should not have happened. Make sure you have pip >= 9.0 to avoid this kind 
of issue, as well as setuptools >= 24.2:

 $ pip install pip setuptools --upgrade

Your choices:

- Upgrade to Python 3.

- Install an older version of Frobulator:

 $ pip install 'frobulator<6.0'

It would be great if you can figure out how this version ended up being
installed, and try to check how to prevent that for future users.

See the following URL for more up-to-date information:

https://i.am.an/url

""")


```

lantz:
Ah, good point! Interesting about ‘pip’ - I guess adding it to the setup params is more useful for clarity or documentation.

lantz:
Haha victory!

Issue #2064, opened 2018-06-11
Sunset python versions <3.5


gizmoguy:
- [x] Disable deb package building for Debian Jessie on gitlab
- [x] Disable unit testing on python 3.4
- [x] Freeze packagecloud repo for Debian Jessie
- [x] Update python package metadata to say we need python >=3.5
- [x] Remove documentation for how to install on Debian Jessie
- [x] Email users/dev mailing lists announcing deprecation
- [x] Bump faucet major versions

gizmoguy:
#2750 disables building on debian jessie

PR #2860, opened 2019-03-15
Sunset python versions < 3.5

gizmoguy:
#2064 

#2372

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,10 +24,6 @@ jobs:
   include:
     - stage: test
       language: python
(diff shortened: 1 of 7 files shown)


gizmoguy:
In #2372 I mentioned we might want to exit early from faucet python programs when python <3.5 is detected.

Do we still want to do this? And if so do we want to add the exit code to ``__main__.py`` ``faucet.py`` ``gauge.py`` ``check_faucet_config.py`` ``fctl.py``?

PR #2883, opened 2019-03-24
packets on stacking ports are marked as external_loop_protect (vlan_pcp=0) even when they're not.

grafnu:
One half of a problem I've been tracking down. I have it distilled down to a test, but haven't identified the solution yet. The problem is that the actions generated for stacking ports on a switch set the vlan_pcp=0, which implicitly makes them considered "external" by the receiving switch. It breaks, then, internal --> external communication across the stack.

If you can point me roughly at where in the code this happens, I'd be most appreciative and can fix it up. Otherwise I'll keep on looking... I just haven't found where the rule actions are constructed yet...

### DIFF
--- a/faucet/valve_flood.py
+++ b/faucet/valve_flood.py
@@ -71,16 +71,20 @@ class ValveFloodManager(ValveManagerBase):
         return list(vlan.flood_ports(vlan.get_ports(), exclude_unicast))
 
     @staticmethod
-    def _build_flood_local_rule_actions(vlan, exclude_unicast, in_port, exclude_all_external):
+    def _build_flood_local_rule_actions(vlan, exclude_unicast, in_port, exclude_all_external, logger):
         """Return a list of flood actions to flood packets from a port."""
-        external_ports = vlan.loop_protect_external_ports_up()
+        external_ports = vlan.loop_protect_external_ports()
         exclude_ports = vlan.exclude_same_lag_member_ports(in_port)
+        if in_port and in_port.number == 1:
+            logger.info('lag_ports ' + str(exclude_ports))
+            logger.info('external_ports ' + str(external_ports))
+            logger.info('loop_protect ' + str(in_port.loop_protect_external))
         if external_ports:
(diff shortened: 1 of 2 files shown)


anarkiwi:
How does this PR relate to https://github.com/faucetsdn/faucet/pull/2884?

grafnu:
It's the same test case as #2884 (external ports on different switches), but a different issue and root cause. I've updated the title of both to be more descriptive.  This does not yet have a proposed fix, just a failing test -- I'm still trying to figure out the right fix.

grafnu:
Ok -- I updated the PR with a working fix for the stack-size-2 case. I wanted to get your feedback before generalizing to the non-size-2 code.

I also don't understand the case where there is no in_port on the flow -- and not sure how to handle it properly.

anarkiwi:
Thanks! The problem description is much clearer. 

We have an overall goal of using the minimum number of flows. Flooding is the most expensive in terms of flows that use a lot of instructions (a lot of output actions). So we attempt to only generate per-port flooding rules when absolutely necessary, and have a lower priority "no port" flow to handle the default case. If a correct per port flow is being discarded in favor of the incorrect default case flow, then the right approach is to fix the logic that causes the per port flow to be discarded (at the moment it compares some actions and output ports, but perhaps it is wrong here). 

A way to debug this further - you can turn on combinatorial flood mode:

combinatorial_port_flood: True

This forces the per-port flow to be used always. If that results in correct behavior, that confirms that the per-port rule is being incorrectly discarded.

grafnu:
Ah, the question/problem wasn't "that the per-port rule is being incorrectly discarded", but rather that I didn't understand what that bit of code was doing (in_port == None) so wasn't sure what to do with it. 

Can you PTAL at the PR fix and let me know if it's headed in the right direction?  It's only done for the stack2 case, and if it's more-or-less correct, I'd proceed to testing/fixing the other cases (n > 2) and properly testing the default flood rule (which hits some weird corner cases with external ports, maybe).

PR #2884, opened 2019-03-24
Configuration reload mistakes external_loop_protect ports as not protected (no longer have vlan_pcp match/actions)

grafnu:
On reload, the pcp match rules get borked.  Traced the problem down to vlan.loop_protect_external_ports_up() not behaving correctly (ports aren't marked as physically up at that point), but haven't figured out the right fix yet.


anarkiwi:
Can you update the PR title/description please with some more detail?

I think you are saying you are fixing a bug where loop protection is not working after a config reload, but it's not clear.

Please also remove the diagnostic info logger (or demote it to logger.debug())

I can see you have a stacking test change where you have only one switch to switch link. We need to cover redundant stacking links, so if you have found a bug with the non-redundant scenario please add a new test case/class for that as required.

anarkiwi:
Also if you can avoid ambiguous terminology (borked) that'd be great. Do you mean, rules are deleted? Changed in some way that are missing ports?

grafnu:
Updated title and tests to check more cases, PTAL.

Issue #2855, opened 2019-03-12
Chewie silently dies if it can't bind to its socket


gizmoguy:
When chewie can't bind its raw socket it just stops logging, It should instead log an error.

anarkiwi:
Verified that misconfiguring Chewie to use a non existent socket (say) causes a crash rather than just silently failing.

PR #2885, opened 2019-03-25
Check for dead threads

anarkiwi:


gizmoguy:
Does this solve and close #2855 ?

PR #2865, opened 2019-03-17
periodic reauth

Bairdo:
closes #2831 

Builds on top of PR #2856 

Requires https://github.com/faucetsdn/chewie/pull/117

PR #2896, opened 2019-04-05
add event notifications for 802.1x

### LINKED ISSUE #2840: Event channel notifications of 802.1x/MAB activity.

Bairdo:
closes #2840 

@grafnu does this suit?
events for auth success, failure, logoff, and port changes of 1x ports. 

I'm wondering if {DOT1X: {PORT_UP: {dpid: 1234, port: 21, ...}}}
would be better than {DOT1X_PORT_UP: {dpid: 1234, port: 21, ...}}

Should add tests.


grafnu:
Yes, seems about right!  I would definitely bundle all the events under one "DOT1X" bucket, since then it greatly simplifies the event processor, which can modularie based off of top-level keys.

@anurag6 -- FYI

Bairdo:
I've only added the tests to Success, Failure, PortStatus. as it is currently checking the order the events occur in. When periodic reauth is in this may need to change, (authentications could occur out of order).

gizmoguy:
Hi @Bairdo I lost my notes from the dev call yesterday. Did you want this PR merged as part of faucet 1.9.2 or aiming for the next release?

Bairdo:
@gizmoguy I didn't at the time. but it should be good to go now (I've fixed up the issue and added some tests)

gizmoguy:
Needs a rebase to add support for chewie 0.0.15 after #2865 was merged

gizmoguy:
This is causing the new periodic reauth tests to fail. I think you might have accidentally undone some of that commit in this PR?

PR #2901, opened 2019-04-08
change chewie handler to accept vlan_name and filter-id

Bairdo:


gizmoguy:
I suspect this commit needs to be part of PR #2865 as that PR is currently failing integration tests (probably due to a python exception)

PR #2905, opened 2019-04-15
extract 8021x base test class. find a free udp port for radius

Bairdo:
Not entirely sure if I'm doing the udp port part right. (What is the ports_sock part?)


anarkiwi:
Looks like file not existing test error?

ports_sock is a local socket used by the test suite to communicate with a process that allocates and manages a pool of known free ports.

Bairdo:
So is it alright to use it to get a UDP port, or is it just for TCP?

PR #2907, opened 2019-04-16
Quick workaround for faucet_config_hash containing invalid characters

anarkiwi:


gizmoguy:
This should already happen in prom client:

https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics.py#L83

Are we generating labels wrong?

anarkiwi:
It looks like they verify label names only at init time. If you redefine labels at runtime (like we do) you dodge the check.

lantz:
@anarkiwi this breaks `faucetagent`

gizmoguy:
@lantz we had to merge this in to get the latest faucet release out. Prometheus labels don't allow "/" in the name and the ``faucet_config_hash`` prom metric was causing prometheus to reject all data from faucet.

Sorry you got caught up in this, it was easier to break faucetagent (since it's still unannounced and we haven't committed to a stable API for it yet) than breaking faucet proper.

We can discuss on the dev call next week the best way forward to make this work again for faucetagent.

lantz:
Ah, OK.

I thought it was nice to have a dict of filename:hash. If the prometheus spec forbids slashes, then that is rather disappointing.

We could make it hash:filename, though that seems a bit backwards.

We could also dump multiple {name:foo hash:bar} label sets, but I'm not sure if that is a good idea either.

lantz:
Possible fix: #2911

PR #2908, opened 2019-04-17
Refactor faucet_dot1x.py

MichaelWasher:
BUGFIX: Remove Default ACL when Authenticated Correctly
Simplify the logging of events inside the dot1x module.
Remove acl_manager dependency from faucet_dot1x.py module
Enforce valve dot1x functions only require port numbers not 'dot1x_port' type.
Remove flow building from handler functions and split into reusable components.
Rename internal variables to conform to Python standard.


anarkiwi:
Thank you - this is so, so much cleaner!

PR #2911, opened 2019-04-19
Update faucet_config_hash_info label format for compatibility with faucetagent

lantz:
The new format is:

`faucet_config_hash_info{config_files='file1,file2,file3', hashes='hash1,hash2,hash3'} 1.0`

This should enable us to fix faucetsdn/faucetagent#12 using faucetsdn/faucetagent#13

### DIFF
--- a/faucet/faucet_metrics.py
+++ b/faucet/faucet_metrics.py
@@ -18,16 +18,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 from prometheus_client import Gauge as PromGauge, Info
 from prometheus_client import Counter, Histogram
 
 from faucet.prom_client import PromClient
-
-
-def quote_label(label):
-    """Return a Prometheus-safe version of a label."""
-    return re.sub(r'[^a-zA-Z0-9_]', '_', label)
 
 
 class FaucetMetrics(PromClient):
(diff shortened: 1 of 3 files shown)


lantz:
I'm not sure if the names are right. Maybe it should be `files` or `filenames` or `paths` or ...??

anarkiwi:
Would it be possible please to do a (manual) test, to make sure a Prometheus instance can scrape FAUCET successfully with this change? It seems we could have "/" in the names again which Prometheus won't like.

lantz:
Hmm, as I understand it the labels should be 'config_files' and 'hashes'. 

Here's what I get when I run it by hand:

```
$ curl -s localhost:9302 | grep config_hash
# HELP faucet_config_hash_info file hashes for last successful config
# TYPE faucet_config_hash_info gauge
faucet_config_hash_info{config_files="/home/openflow/faucetagent/faucet.yaml",hashes="b2a3e0dd05c22b776482bc241dc1d1f6a4f5a3ed072cc5055e9cbee523c6d186"} 1.0
# HELP faucet_config_hash_func algorithm used to compute config hashes
# TYPE faucet_config_hash_func gauge
faucet_config_hash_func{algorithm="sha256"} 0.0
```

Does that look OK?

Also is there an integration test to verify that prometheus scrapers can read FAUCET's prometheus output?

anarkiwi:
Ahhh right, previously the issue was that filenames were the label names, not the values. Thanks!

PR #2964, opened 2019-05-22
Enable lacp_passthrough feature

grafnu:
This is a close-but-not-quite-ready PR that implements a lacp_passthrough feature (similar to things on Cisco and other switches). There's still excess debugging info and other things to clean up, but I wanted to get a high-level read on the solution, and or the appropriate fix for a (already present) reconfiguration bug, at [line 1495 of valve.py](https://github.com/faucetsdn/faucet/compare/master...grafnu:lacpfix?expand=1#diff-3947abbee6fbf8fd826cfc049a25dcdfR1495). If all is well then I'll finalize w/ cleanup etc...


anarkiwi:
Looks like a useful feature - can the bug fix be separated out into its own PR, first please with a test so we can address that separately?

grafnu:
Bug fix submitted as https://github.com/faucetsdn/faucet/pull/2972. Will clean up this PR next.

grafnu:
This PR has been cleaned up and ready for review. PTAL.

PR #2969, opened 2019-05-22
Update copyright for 2019

cglewis:


gizmoguy:
Halfway through 2019 seems like a good a time as any to update the dates :)

PR #2979, opened 2019-05-27
ensure config_hash metric exists even if parsing fails

cglewis:
this is needed for the `faucetagent` specifically for using the `CONFIG_STAT_RELOAD` feature.


lantz:
This is interesting - interestingly enough my agent test generated this situation by writing an empty config.

PR #2988, opened 2019-05-29
Make pip cache easier to use for docker test runs.

gizmoguy:
* Recommend using separate cache directory (rather than the user one)
  * Remove need to prime the cache before use

### DIFF
--- a/docker/runtests.sh
+++ b/docker/runtests.sh
@@ -36,9 +36,8 @@ fi
 
 if [ -d /var/tmp/pip-cache ] ; then
   echo Using pip cache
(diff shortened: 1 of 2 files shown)


lantz:
Will this allow local caching pip packages locally so that it doesn't have to download them every time (as it seems to do, very slowly usually, on my system?)

gizmoguy:
@lantz yes this will.

The current pip cache system assumes you "prime your cache" before you start your docker which most users won't do as it potentially will clobber their pre-existing pip cache.

This change will fix that by causing the docker itself to automatically populate and keep a separate isolated pip cache up to date between runs.

PR #2992, opened 2019-06-02
Allow fixed or random ordering for port_map

lantz:
Experimenting with a simpler and more incremental approach.

Adds a parameter `--order 0,1,2,3 | random`
to the integration tests to allow random or predetermined ordering for `port_map`.

By default, it is `random` which uses a random shuffle of `0,1,2,3`.

This allows tests to exercise hardware or software switch ports in a non-canonical order.

### DIFF
--- a/clib/clib_mininet_test_main.py
+++ b/clib/clib_mininet_test_main.py
@@ -210,13 +210,13 @@ def check_dependencies():
     return True
 
 
-def make_suite(tc_class, hw_config, root_tmpdir, ports_sock, max_test_load):
+def make_suite(tc_class, hw_config, root_tmpdir, ports_sock, max_test_load, port_order):
     """Compose test suite based on test class names."""
     testloader = unittest.TestLoader()
     testnames = testloader.getTestCaseNames(tc_class)
     suite = unittest.TestSuite()
     for name in testnames:
-        suite.addTest(tc_class(name, hw_config, root_tmpdir, ports_sock, max_test_load))
+        suite.addTest(tc_class(name, hw_config, root_tmpdir, ports_sock, max_test_load, port_order))
     return suite
 
 
@@ -370,7 +370,7 @@ def max_loadavg():
 
 
 def expand_tests(module, requested_test_classes, excluded_test_classes,
-                 hw_config, root_tmpdir, ports_sock, serial):
+                 hw_config, root_tmpdir, ports_sock, serial, port_order):
     sanity_test_suites = []
     single_test_suites = []
     parallel_test_suites = []
@@ -388,7 +388,7 @@ def expand_tests(module, requested_test_classes, excluded_test_classes,
                 continue
             print('adding test %s' % test_name)
             test_suite = make_suite(
-                test_obj, hw_config, root_tmpdir, ports_sock, max_loadavg())
+                test_obj, hw_config, root_tmpdir, ports_sock, max_loadavg(), port_order)
             if test_name.startswith('FaucetSanity'):
                 sanity_test_suites.append(test_suite)
             else:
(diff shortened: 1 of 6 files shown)


lantz:
Note this is work in progress. I decided to try something very simple and then make sure it passes on hardware and software switch tests.

lantz:
Ah good catch.

gizmoguy:
This needs to be run against real hardware before merging.

anarkiwi:
Just checking in - what is the current state of testing on hardware for this change?

gizmoguy:
> Just checking in - what is the current state of testing on hardware for this change?

My hardware test bed was shutdown due to my AC failure. I'm just booting it up now to start testing this PR

gizmoguy:
This does not pass on my hardware test bed, it's failing because it doesn't like how I've cabled my switch under test:

```
test_portmap (mininet_tests.FaucetSanityTest) ... DP: of_dp_desc_stats{dp_desc="cisco-9300-2:sw1",dp_id="0xf87b2053de80",dp_name="faucet-1",hw_desc="C9300-24UX",mfr_desc="Cisco Systems, Inc.",serial_num="FCW2140G00F",sw_desc="Cisco IOS Software [Gibraltar], Catalyst L3 Switch Software (CAT9K_IOSXE), Experimental Version 16.10.20180930:011424 [v1610_throttle-/nobackup/mcpre/BLD-BLD_V1610_THROTTLE_LATEST_20180930_010715 167]| openvswitch 2.1"} 2.73207707033216e+014
port_map: {'port_1': 4, 'port_2': 3, 'port_3': 2, 'port_4': 1}
verifying cabling for port_1: host enp1s0f3 -> dp 4
DP 4 is Te1/0/4, at 1000 mbps
verifying cabling for port_2: host enp1s0f2 -> dp 3
DP 3 is Te1/0/3, at 1000 mbps
verifying cabling for port_3: host enp1s0f1 -> dp 2
DP 2 is Te1/0/2, at 1000 mbps
verifying cabling for port_4: host enp1s0f0 -> dp 1
DP 1 is Te1/0/1, at 1000 mbps
FAIL
```

lantz:
Huh, I wonder what that means - the error message doesn't really provide any information unfortunately. Could you include your config?

gizmoguy:
Here's my config:

```
brad@faucet-hw-tester:~$ cat /etc/faucet/hw_switch_config.yaml 
hw_switch: True
# Map ports on the hardware switch, to physical ports on this machine.
# If using a switch with less than 4 dataplane ports available, run
# FaucetZodiac tests only. A 4th port must still be defined here and
# must exist, but will not be used.
dp_ports:
  1: enp1s0f0
  2: enp1s0f1
  3: enp1s0f2
  4: enp1s0f3
# Hardware switch's DPID
dpid: 0xf87b2053de80 #cisco
#dpid: 0x120677c423f80 #aruba
#dpid: 0xeccd6def54fe #at
hardware: 'CiscoC9K'
# Port on this machine that connects to hardware switch's CPN port.
# Hardware switch must use IP address of this port as controller IP.
cpn_intf: eno1
# There must be two controllers configured on the hardware switch,
# with same IP (see cpn_intf), but different ports - one for FAUCET,
# one for Gauge.
of_port: 6653
gauge_of_port: 6654
# If you wish to test OF over TLS to the hardware switch,
# set the following parameters per Ryu documentation.
# https://github.com/osrg/ryu/blob/master/doc/source/tls.rst
# ctl_privkey: ctl-privkey.pem
# ctl_cert: ctl-cert.pem
# ca_certs: /usr/local/var/lib/openvswitch/pki/switchca/cacert.pem
```

From what I can tell from the output of FaucetSanityTest, my ports are wired to match my config file:

```
host enp1s0f3 -> dp 4
host enp1s0f2 -> dp 3
host enp1s0f1 -> dp 2
host enp1s0f0 -> dp 1
```

But I think that your randomization code has run and reordered the ports to be backwards:

`port_map: {'port_1': 4, 'port_2': 3, 'port_3': 2, 'port_4': 1}`

lantz:
The port map is supposed to be randomized, so that should be fine. The config also looks OK, so I am not sure what is going on. I will test it again on my hardware testbed.

Can you confirm that the OpenFlow ports are assigned appropriately on your switch as well?

lantz:
AC failure - that is a real hardware failure!! D:

anarkiwi:
I ran this PR on my virtual hardware testbed and it passed.

lantz:
Any success (or failure) on the hardware testbed?

PR #2998, opened 2019-06-05
Fix crash when optional include file does not exist.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2952

PR #3011, opened 2019-06-13
LLDP src validation

anurag6:
Validate LLDP peer mac address against a configurable parameter.


anarkiwi:
Please add a config unit test for this (that the str is a valid MAC address).

anurag6:
Checking whether configured mac is a src mac or not.

PR #3012, opened 2019-06-13
Making LACP resp rate configurable

anurag6:
'lacp_resp_interval' regulates the min time between each response to a LACP packet is sent out.


anarkiwi:
Please add a config unit test for this (making sure the value is from a valid range - eg handling 0 or an implausibly high number).

anurag6:
Added a few negative testcases to check for negative/zero and very high values.

gizmoguy:
@anurag6 could you rebase this after we merged your other PR. If you get it rebased in the next 3 hours we can guarantee this gets in this weeks release.

anurag6:
@gizmoguy done merging with latest.

anarkiwi:
The comments for the unit tests are not correct (they say test for valid DP ID), but OK for now.
Please open a clean up PR for the next round.

PR #3027, opened 2019-06-19
Removing test causes unrelated failure

grafnu:
Hoping somebody can shed some light on this.  I've been working with a test and anything I do to it seems to cause an unrelated failure in another test. This PR simply removes one test, and then something else fails.

Am I missing something here?

I'm removing FaucetTaggedExtLoopProtectTest, and then FaucetUntaggedBGPIPv6RouteTest fails!

Failure is consistent -- happens every time but I don't understand at all why.

### DIFF
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -5180,37 +5180,6 @@ class FaucetTaggedBroadcastTest(FaucetTaggedTest):
         self.verify_no_bcast_to_self()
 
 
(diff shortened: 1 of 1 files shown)


anarkiwi:
You might not have merged master recently which contains a fix for an exabgp upgrade.

gizmoguy:
This passed fine on travis, I think @anarkiwi is correct that you got bitten by the new version of ExaBGP same as we did #3018

PR #3030, opened 2019-06-20
Move loop_protect_external logic to vlan table

grafnu:


grafnu:
The existing tests aren't catching some regressions (manual inspection turned up some problems), so going to fix those then re-submit.

grafnu:
Never-mind -- I was confusing myself. I inspected the flow tables and they look right and so the test is covering them properly.

PR #3060, opened 2019-06-21
First pass at a "silence" sanity check; must hear no unexpected packe…

anarkiwi:
…ts in test environment where all hosts are deconfigured.


lantz:
I like this idea. I assume that disabling ipv6 in the root namespace is OK? Presumably doing so in the host namespaces will only be for that test.

PR #3085, opened 2019-07-02
WIP: Verify host/switch sort order

lantz:
We verify that `topo.hosts()` returns the hostnames in expected lexical order.

We also verify that `topo.hosts(sort=False)` returns the hostnames in `addHost()` order.

We also verify that `hosts_named_ordered()` is equivalent to:

    [self.net[hostname] for hostname in self.topo.hosts()]

and likewise for `switches_named_ordered()`.

Note that they could potentially diverge in the future since `hosts_named_ordered()` is currently using string rather than lexical order, so `u11` < `u2`.

### DIFF
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -1749,6 +1749,19 @@ class FaucetSanityTest(FaucetUntaggedTest):
                 self.tcpdump_rx_packets(tcpdump_txt, 0),
                 msg='got unexpected packet from test switch: %s' % tcpdump_txt)
 
+    def test_name_sorting(self):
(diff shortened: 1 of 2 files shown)


lantz:
This is interesting and unexpected. The unit test works fine for me locally but seems to be failing on travis.

anarkiwi:
As we currently have a working solution, I I propose we postpone further investigation unless there is an alternative?

lantz:
I am still thinking about this, hence WIP. I would like to know why things aren't working the way I expect, but I'm fine closing it for now and reopening later when I get the chance to look at it more. I am thinking that... probably the solution belongs in the topo.

PR #3112, opened 2019-07-23
Support TFM for OVS (2.11.0 or later) with "TFM Open vSwitch" hardware type

anarkiwi:


faucet/valve.py
```diff
@@ -1729,6 +1745,7 @@ class NoviFlowValve(Valve):
     'Netronome': OVSValve,
     'NoviFlow': NoviFlowValve,
     'Open vSwitch': OVSValve,
+    'TFM Open vSwitch': OVSTfmValve,
```
gizmoguy: I'm just wondering if "Open vSwitch TFM" might make more logical sense. It would follow our "Generic" and "GenericTFM" we already have.

anarkiwi: Done.

PR #3119, opened 2019-07-28
Verify that switch works after several disconnections

lantz:
First we change the DPID to something incorrect and then wait for FAUCET to disconnect from the switch several times.

Then we restore the correct DPID, reload the config, and verify that `untagged_test()` still works.

Notes: 

Initially I was dissatisfied with this approach for a couple of reasons, but now I think maybe it isn't so bad. Most of the delay seems to be from waiting for the switch to reconnect and for performing the connectivity test. Even if we did this in another way, we would still have to wait for the switch to reconnect, which is a switch configuration, and we would still want to run the connectivity test. We could try a simpler/faster test however (e.g. just do the ping test.)

### DIFF
--- a/clib/mininet_test_base.py
+++ b/clib/mininet_test_base.py
@@ -2308,16 +2308,16 @@ dbs:
         exp_re = re.compile(exp)
         with open(log_name) as log_file:
             return [log_line for log_line in log_file if exp_re.match(log_line)]
-        return []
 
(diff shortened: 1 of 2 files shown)


anarkiwi:
https://github.com/faucetsdn/faucet/blob/master/clib/mininet_test_base.py#L2313

You could also use an existing test helper, wait_until_matching_lines_from_file()

lantz:
Aha, good idea. ;-)

PR #3122, opened 2019-07-29
FaucetBadFlowModTest: Send bad flow mods to switch

lantz:
We generate some questionable flow mods and send them to the switch.
Then we verify that hosts can still ping each other.

This seems to result in some (presumably correct) OpenFlow errors from the switch. OvS in particular (on my HW testbed) doesn't seem to like the MPLS actions. On the other hand it seems perfectly OK with bogus port and table numbers.

Unfortunately posting via `ofctl` is very slow, but it does seem to work. For that reason, we currently only send 10 flow mods, but we could send more.

Some things to think about:

- Do we want to try bogus flow adds/deletes as well as modify operations?
- Do we want to try to generate specific errors for TFM switches?
- Is this valuable? Are there other bad OF messages we might want to add?

### DIFF
--- a/clib/mininet_test_base.py
+++ b/clib/mininet_test_base.py
@@ -360,8 +360,9 @@ class FaucetTestBase(unittest.TestCase):
     def hostns(self, host):
         return '%s' % host.name
 
-    def tearDown(self):
-        """Clean up after a test."""
+    def tearDown(self, ignore_oferrors=False):
+        """Clean up after a test.
+           ignore_oferrors: return OF errors rather than failing"""
         if self.NETNS:
             for host in self.net.hosts[:1]:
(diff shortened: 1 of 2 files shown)


gizmoguy:
I should test this on a TFM switch to see what happens.

anarkiwi:
This looks like a great start.

Overriding assertFalse is a bit of a concern:

def assertFalse(self, expr, msg=None):

Would you be able to add a test attribute (like we do for whether meters are supported, etc) so that tearDown() knows that errors are OK?

lantz:
I added a flag to `tearDown()` to allow us to return the OF errors rather than failing. We can also potentially test for specific errors.

Issue #2672, opened 2018-12-09
intervlan routing with multiple dp connected with trunks does not work 


kstest1:
I have 3 dp: s1, s2, s3.  Configured vlans: 100 and 200. Configured router on vl100 and vl200. 
**Switch connections:**
s1, port 20 (of port1 -see config)  connects to s2, port 20  (of port1) . vlan 100 and 200 allowed with tagging.
s2, port 21 (of port2)  connects to s3, port 21. (of port2) vlan 100 and 200 allowed with tagging.
**Host connections:**
On s1, port 22  (of port3) , there is host h1 in native vlan 100 ( ip 10.0.100.51/24, d.g. 10.0.100.254) 
On s3, port 22  (of port3) , there is host h2 in native vlan 200.( ip 10.0.200.52/24, d.g. 10.0.200.254) 
On s3, port 20  (of port1) , there is host h3 in native vlan 100.( ip 10.0.100.53/24, d.g. 10.0.100.254) 
**Result:**
h1 can ping h3 in same vlan. 
h2 can ping h3 in different vlan ( they are on s3 switch).
h1 can NOT ping h2 in different vlan ( they are on s1 /s3  switches connected via s2).

**Reason:** 
When h2 pings h1, s3 needs to know arp info for h1 in vlan 100. arp is generated from faucet and sent from s3 , port 21  to s2, port21. This packet is dropped on s2 as came from faucet vip dl address. Even if it is allowed by config, return packet ARP will be taken by  s1 as reply go to same mac address and s3 will not know mapping. 

**Config** 
(   test with  hardware: "GenericTFM", same with other hw) 
```

test#more /etc/faucet/faucet.yaml 
include:
   - acls.yaml


vlans:
    office:
        vid: 100
        description: "office network"
        faucet_mac: "0e:00:00:00:10:01"
        faucet_vips: ['10.0.100.254/24', '2001:100::1/64', 'fe80::c00:00ff:fe00:1001/64']
    guest:
        vid: 200
        description: "guest network"
        faucet_mac: "0e:00:00:00:20:01"
        faucet_vips: ['10.0.200.254/24', '2001:200::1/64', 'fe80::c00:00ff:fe00:2001/64']

routers:
    router-office-guest:
        vlans: [office, guest]

dps:
    sw1:
        dp_id: 0x1
        hardware: "GenericTFM"
        interfaces:
            3:
                name: "if22"
                description: "if22-offvl"
                native_vlan: office
            1:
[...]

PR #3127, opened 2019-08-04
Enable inter-VLAN routing with stacking

mab68:
- Enable the ability to use stacking with inter-VLAN routing (#2672 ).
- drop_spoofed_faucet_mac should be set to false to allow a routed packet enter the eth_src table on another switch.
- A valve receiving packets will call other valves in the stack to also learn the host.


anarkiwi:
Looks like valve_host is 0.01 lint score short?

PR #3132, opened 2019-08-05
add basic adapter for faucetagent

cglewis:


lantz:
Oh interesting; do you want to explain a bit about what this does? I'm not familiar with adapters - what do they do?

Also, is there a test for this?

PR #3161, opened 2019-08-12
Fix recovery from a bad config.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/3131

PR #3163, opened 2019-08-12
Fix unhandled exception when LLDP TTL out of range

mab68:


anarkiwi:
Looks like needs some linting.

PR #3182, opened 2019-08-15
ebtables needs to be run before the controller starts to avoid learning test host MAC addresses

anarkiwi:


lantz:
I agree! 

I pulled it out as well in #37

PR #3193, opened 2019-08-20
lacp health should factor into stack root health

anarkiwi:


anarkiwi:
Also fixes https://github.com/faucetsdn/faucet/issues/3191

lantz:
Oh I see. This seems fine - I think it might also be possible to use `port_map` or the switch port order that `dpid_ports` returns.

PR #3195, opened 2019-08-21
WIP: Cache faucet_config_table_names to avoid delay

lantz:
If we repeatedly scrape these values and wait for tables that will never appear, it can take a long time (around 30 seconds). Since `faucet_config_table_names` seems fairly static, we should be able to cache it.

My quick/local tests: 
- `FaucetSanityTest` (master): 201s
- `FaucetSanityTest` (prometheus-cache): 90s

There may be better ways of doing this, but it seems like we should be able to save some time.

### DIFF
--- a/clib/mininet_test_base.py
+++ b/clib/mininet_test_base.py
@@ -375,7 +375,9 @@ class FaucetTestBase(unittest.TestCase):
                     'dump-flows', 'dump-groups', 'dump-meters',
                     'dump-group-stats', 'dump-ports', 'dump-ports-desc'):
                 switch_dump_name = os.path.join(self.tmpdir, '%s-%s.log' % (switch.name, dump_cmd))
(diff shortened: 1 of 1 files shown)


lantz:
Hmm, this build passed:
https://travis-ci.org/lantz/faucet/jobs/575122543

anarkiwi:
So this is definitely a thing.

It should be safe to wait for the first table ID variable to appear, and then not retry scrapes for all the others.

There is always a VLAN table, so you could retry waiting for that, then not retry for all the other tables.

anarkiwi:
It looks like this PR is stale (thank you for the inspiration to speed up table ID resolution, I addressed that slightly differently). It would be good to explore whether caching could speed up other cases (e.g. avoid ofctl calls multiple times per second).

PR #3252, opened 2019-09-23
Allow FAUCET to optionally, automatically revert bad configs itself

anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/2719.

Setting FAUCET_CONFIG_AUTO_REVERT enables the feature.


gizmoguy:
It would be good to add documentation for this new env var to configuration.rst:

https://github.com/faucetsdn/faucet/blob/master/docs/configuration.rst#environment-variables

Issue #3176, opened 2019-08-14
Feature request: generic coprocessor function


cglewis:
**Scenario:**

Would like the ability to have some specified traffic (via ACL) have to go first through some NFV process then back out the intended destination. To do this, it could be easily driven by having special VLAN tags that are dedicated for the purposes of forcing the traffic through this intermediate port/destination before coming back out and going to its originally intended destination.

To make this least cumbersome for having to keep track of special VLAN tags for this purpose (or not clobbering other purposes, such as Chewie) it would be nice to be able to have a label/tag (string) in the config that could represent the intent of needing a special VLAN without explicitly hardcoding a specific number dedicated for it. Faucet would then take the label/tag and translate it into an unused VLAN number (based on parsing the config) that it can use for those special needs. 

Since later down the road the config could be changed to include the VLAN number that was picked by Faucet, Faucet should be able to dynamically allocate this VLAN number every time the config gets reloaded in case the previously allocated VLAN is now in use by the config.  This could of course be an optional configuration field or perhaps environment variable if it isn't desired to always have to run this logic in the controller.

**Possible use cases:**

- Throttling
- DPI
- Reshaping
- etc.

PR #3258, opened 2019-09-26
Initial coprocessor implementation (can inject packets into pipeline and/or output directly to DP ports)

anarkiwi:


gizmoguy:
#3176

faucet/port.py
```diff
@@ -168,7 +176,6 @@ def __init__(self, _id, dp_id, conf=None):
         self.name = None
         self.native_vlan = None
         self.number = None
-        self.op_status_reconf = None
```
gizmoguy: Did you mean to delete this?

anarkiwi: Yes that is a vestigial duplicate not referenced anywhere.

PR #3261, opened 2019-09-29
Allow MST flooding from NoReflection algorithm with wounded stack.

grafnu:
Provisional fix to NoReflection algorithm that allows a system to operate with calculated-MST flooding. Alternative solution would be to create another flood manager with this behavior. Still needs tests.

### DIFF
--- a/faucet/valve_flood.py
+++ b/faucet/valve_flood.py
@@ -333,6 +333,9 @@ class ValveFloodManager(ValveManagerBase):
 class ValveFloodStackManagerBase(ValveFloodManager):
     """Base class for dataplane based flooding on stacked dataplanes."""
 
+    # By default, no reflection used for flooding algorithms.
+    _USES_REFLECTION = False
+
     def __init__(self, logger, flood_table, pipeline, # pylint: disable=too-many-arguments
                  use_group_table, groups,
                  combinatorial_port_flood, canonical_port_order,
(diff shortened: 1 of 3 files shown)


grafnu:
Updated with unit test to demonstrate functionality change. PTAL.

gizmoguy:
Hi @grafnu 

Apologies but looks like we already had some pylint problems in tests/unit/faucet/test_valve_stack.py and your changes here just pushed us over the bad threshold.

I think the easiest ones to fix will be:

```
/faucet-src/tests/unit/faucet/test_valve_stack.py:342:8: C0103: Variable name "dp" doesn't conform to snake_case naming style (invalid-name)
/faucet-src/tests/unit/faucet/test_valve_stack.py:391:8: C0103: Variable name "dp" doesn't conform to snake_case naming style (invalid-name)
/faucet-src/tests/unit/faucet/test_valve_stack.py:860:8: C0103: Variable name "dp" doesn't conform to snake_case naming style (invalid-name)
/faucet-src/tests/unit/faucet/test_valve_stack.py:899:8: C0103: Variable name "dp" doesn't conform to snake_case naming style (invalid-name)
```

Just add a **# pylint: disable=invalid-name** next to these lines like so:

`dp = self.valves_manager.valves[self.DP_ID].dp # pylint: disable=invalid-name`

And hopefully after that you should be passing on travis

grafnu:
Well then. Seems like I somehow managed to make the *unit* test flaky, so I
need to address that first.  Lint is easy :-), although I was actually
renaming the variable dp_obj rathre than the pylint supression.

grafnu:
Fakyness and some other issues addressed separately in PR https://github.com/faucetsdn/faucet/pull/3273 -- once that's submitted, I'll update this to match and re-request review.

gizmoguy:
#3273 is in now, thanks. Feel free to rebase this one and we'll take another look.

PR #3271, opened 2019-10-02
test_passthrough accesses wrong dp for lacp

snak1219:
Otherwise, test fails by hitting a *key error*.


snak1219:
this is for **FaucetStringOfDPLACPUntaggedTest**

tests/integration/mininet_tests.py
```diff
@@ -7447,6 +7447,7 @@ def test_passthrough(self):
         interfaces_conf[dst_port]['loop_protect_external'] = True
         interfaces_conf[dst_port]['lacp'] = 2
         interfaces_conf[src_port]['loop_protect_external'] = True
+        interfaces_conf = conf['dps']['faucet-2']['interfaces']
         interfaces_conf[fail_port]['loop_protect_external'] = True
         interfaces_conf[end_port]['loop_protect_external'] = True
```
snak1219: These interfaces are the "peer switch" ports which should be accessed from faucet-2, not faucet-1.

PR #3284, opened 2019-10-09
Use max_lldp_lost properly

grafnu:
Fixes bug to enable easy workaround for https://github.com/faucetsdn/faucet/issues/3281


anarkiwi:
Please add a test that shows that the parameters are used correctly.

anarkiwi:
Thanks that's better, and it shows the parameter can be parsed, but it doesn't show that the parameter is acted on?

grafnu:
Updated with a comment explaining how the test tests behavior... PTAL.

anarkiwi:
I don't see how the max lldp port option is tested still?

grafnu:
Updated with more comments, ptal. I know the behavioral test is there, but
it is admittedly buried somewhat down in the testing library (the use of
LLDP beacons). Not sure if there's a better way...

anarkiwi:
I still don't see how the parameter fix works, though? I'm looking for a test that shows, that you can set a particular number (like 100), and that we wait the appropriate number of lost packets.

grafnu:
I made the number of packets it was waiting for more explicit, so the bounds should be clear. First it's testing the default behavior (3), and then sets the value to 100 and tests the new behavior...

PR #3292, opened 2019-10-18
Change the output format for gauge text flow table stats

clorier:
This branch failed travis on my github and on my desktop, but passed when I accidentally pushed to my old github https://github.com/KitL/faucet/runs/264929690
The failure isnt obviously connected with the changes I have made.
So I'm not sure what is going on with it at the moment, but I figured if I make the PR I will get a known good environment for running the tests.

    This is a breaking change.

    The old format involved severe hackiness to make it manageable
    This format instead will just dump each OFStatsReply message
    straight into its own file.

    The filename will be {datapath_name}-flowtable-{timestamp}.json
    The directory these files will be stored can be configured using
    the "path" key in the configuration for a text db

    The tests and the default gauge.yaml have been changed accordingly

    The gauge unit tests now create a temporary directory rather than
    a temporary file, then a file within that directory is created
    (currently called temp_filename) for use with the port_stats tests.

    It is probably a good idea that port stats change to use a similar
    format in future.

### DIFF
--- a/clib/mininet_test_base.py
+++ b/clib/mininet_test_base.py
@@ -192,10 +192,11 @@ class FaucetTestBase(unittest.TestCase):
             self.tmpdir, 'gauge-ports.txt')
         self.monitor_state_file = os.path.join(
             self.tmpdir, 'gauge-state.txt')
-        self.monitor_flow_table_file = os.path.join(
-            self.tmpdir, 'gauge-flow.txt')
+        self.monitor_flow_table_dir = os.path.join(
+            self.tmpdir, 'gauge-flow')
         self.monitor_meter_stats_file = os.path.join(
             self.tmpdir, 'gauge-meter.txt')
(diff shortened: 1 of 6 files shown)


anarkiwi:
Looks like a stickler lint change?

gizmoguy:
WIP, needs more thought regarding multi-part messages.

clorier:
So what I have done with multipart messages is the simplest thing possible. That is, I write them to separate files. I just added a way to add an increment to a filename if it gets two messages in the same second.

So a programme that interacts with the files should check for the OFPMPF_REQ_MORE flag to see whether there are additional files it should read, and look for something with either the incremented filename or something that arrived within the next few seconds.

clorier:
Codecov is complaining about files I havent touched, so I am leaving that.

Otherwise should be ready to go.

PR #3299, opened 2019-10-20
FAUCET should handle a bad initial config if auto revert is enabled.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/3289.

PR #3317, opened 2019-10-31
Add more tests for expected loop_protect_external

grafnu:


tests/integration/mininet_tests.py
```diff
@@ -6979,6 +6979,17 @@ def verify_tunnel_established(self, src_host, dst_host, other_host, packets=3):
             '%s: ICMP echo request' % other_host.IP(), tcpdump_text
         ), 'Tunnel was not established')
 
+    def verify_one_broadcast(self, from_host, to_hosts):
+        self.assertGreater(len(to_hosts), 1, 'Testing only one ext host is not useful')
+        received_broadcasts = []
+        for to_host in to_hosts:
+            try:
+                self.verify_broadcast(hosts=(from_host, to_host), broadcast_expected=False)
+            except Exception:
```
anarkiwi: What exception are you expecting here?

grafnu:
I changed it to AssertionError and added a comment. It's basically catching
the exception of when a broadcast is received (since expected=False), and
counting them.  Other way would be to refactor the test library code to be
able to (optionally) return the broadcast state, rather than force an
assertion.

anarkiwi:
That would be better (returning a bool). Catching Exception is generally not good because it will catch potentially other issues, and even catching assertionError may cause confusion if the assertError is being raised by some other check than than the one you're interested in (broadcast behavior).

grafnu:
Ok, updated with proper bool return. Please take a look when you get a chance!

PR #3318, opened 2019-10-31
Bigger icmp

### LINKED ISSUE #3316: Allow faucet to receive ICMP messages of larger sizes
Apparently some devices require larger ICMP messages than faucet currently supports:

https://lists.geant.org/sympa/arc/faucet-users/2019-10/msg00050.html

I propose raising our ICMP size limit to 128b of payload and if this doesn't fix these devices we can implement a configurable value for setting how large of an ICMP packet you want faucet to be able to process.

gizmoguy:
* Standardize ICMP packet truncation for v4 and v6 to 128 bytes.

 * Introduce new MAX_PACKET_IN_SIZE field for truncating things that aren't ICMP that we don't know the actual length of

fixes #3316


tweippert:
I tested the patch on the environment where i had the issues, it works for both Cisco ASR 1001-X and VMWare ESX Isolation Tests. So the higher icmp payload patch works as expected:

ASR-1001-X#ping 10.8.64.129 size 100
Type escape sequence to abort.
Sending 5, 100-byte ICMP Echos to 10.8.64.129, timeout is 2 seconds:
!!!!!
Success rate is 100 percent (5/5), round-trip min/avg/max = 5/5/6 ms

PR #3323, opened 2019-11-01
Help re: why does this fail?

grafnu:
I'm trying to write some tests for what happens when certain ports go up/down in a configuration, but I'm having a really hard time figuring out how to incite the desired behavior. There's the "set_port_down" function, but I can't seem to get it to work for anything other than the first dp (self.dpids[0]) -- that works, but when I try it with the other dp it fails.  Any help understanding how to bring a port down during a test would be greatly appreciated.

The hard-coded port 5 shouldn't matter, and doesn't seem to cause any problems. In this example test case I don't actually care which port is down, just that I want to set a port down on the second switch.

### DIFF
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -7508,7 +7508,12 @@ class FaucetSingleStackStringOfDPExtLoopProtUntaggedTest(FaucetStringOfDPTest):
             use_external=True)
         self.start_net()
 
-    def test_untagged(self):
+    def test_port_down(self):
+        """Basic port up/down functionality works"""
+        self.set_port_down(5, self.dpids[0])
+        self.set_port_down(5, self.dpids[1])
+
+    def x_test_untagged(self):
         """Host can reach each other, unless both marked loop_protect_external"""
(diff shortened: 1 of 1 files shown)


gizmoguy:
Alright, it took me a bit of digging, but I think I finally cracked this one.

You hit a bug in set_port_down/set_port_status/wait_port_status/scrape_prometheus_var. Historically the integration tests only ever dealt with a single datapath so some functions still make this assumption. These functions were assuming every datapath name was "faucet-1" but for FaucetStringOfDPTest datapaths are named "faucet-**N**". I implemented a bit of a hack to test my theory here, hopefully you might be able to come up with something slightly cleaner:

https://github.com/gizmoguy/faucet/commit/18f2919a3c76f1ed4e31218585f043b37df8303a 

I also rewrote the test to use the port_map to lookup the OF port number for port 5 which will work with the random port ordering code that's in there:

https://github.com/gizmoguy/faucet/commit/0ec753769dca8ec5d288066b719ce678defb2e7c

I'll close this PR for now, get back in touch if this fix doesn't work for you or you want some more information on the problem.

PR #3327, opened 2019-11-04
Add lacp_collect_and_distribute port config option

samribeiro:
Allow a multi root stack to aggregate its LACP ports to non MC-LAG devices.

Normally Faucet sets one port (on the root stack switch) to Collecting and Distributing (Active) and the others to not Collecting and not Distributing (Standby). This works fine with MC-LAG enabled devices where different devices establish aggregation via multiple ports as it they were one device. However, enabling this for non MC-LAG devices is just a matter of overriding the need to Standby. This knob forces all LACP configured ports to Collecting and Distributing.

Because enabled ports will all state that they desire to collect traffic, they will receive traffic, however this traffic will always be dropped on non root stack switch.

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -735,6 +735,12 @@ configuration.
                 return 0
         return 1
 
+    def lacp_collect_and_distribute(self, port):
+        """Return 1 if LACP should advertise collect and distribute on this port."""
+        if port.lacp_collect_and_distribute:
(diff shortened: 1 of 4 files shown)


anarkiwi:
Can we get a minimal unit test for this please?

samribeiro:
Certainly, will submit them tomorrow.

PR #3341, opened 2019-11-14
Update lacp state machine

grafnu:


gizmoguy:
LGTM, because this changes definition of the ``port_lacp_state`` prom variable (exposes more information that we did before). I will need to announce breaking change in the release notes, however I don't think this will annoy enough people to warrant incrementing major version number of introducing an entirely new prom variable.

Issue #3347, opened 2019-11-18
Faucet always cold restarts when a "router" is defined


gizmoguy:
From @mwutzke:

> If I remove the routers section, I no longer get the cold start operation - I get the expected warm start operation, for a simple port name change.
> 
> I suspect the issue is in this logic:
> 
> >        elif new_dp.routers != self.routers:
> >            logger.info('DP routers config changed - requires cold start')
> 
> I *think* - need to check, that a 'deep' compare needs to be used - but I've not verified that.

Possibly related to #3345

mwutzke:
The vip_map_by_ipv attribute of faucet.Router contains PyTricia objects, and the 'str' representation of these objects renders the 'id()' of the object. eg:

```
 "('vip_map_by_ipv', OrderedDict([(4, <pytricia.PyTricia object at 0x7fc30da8cf50>), (6, <pytricia.PyTricia object at 0x7fc30da8ce90>)]))"
```

As such the 'hash' will be different for 2 logically identical objects, resulting in the 'router config changed' path being executed.

PR #3350, opened 2019-11-19
router.py: Add 'repr' implementation for vip_map_by_ipv (#3347)

mwutzke:
To create ia 'hash'able Router class, the PyTricia
tree needs to be unrolled (the 'id()' based 'repr'
does not allow for comparison of equivalent objects)


mwutzke:
Ill address the issues and resubmit a new pull request

PR #3369, opened 2019-11-29
Report config hash change

grafnu:


clib/mininet_test_base.py
```diff
@@ -241,6 +246,26 @@ def _enable_event_log(self, timeout=None):
...
+                for event_str in events:
+                    event = json.loads(event_str)
+                    event_id = event['event_id']
+                    if event_id <= self.prev_event_id:
+                        continue
+                    self.prev_event_id = event_id
+                    try:
+                        if match_func(event):
+                            return event
+                    except Exception:
```
anarkiwi: Please replace with the actual expected exception.

grafnu: Updated.

PR #3376, opened 2019-12-05
Add txerror to prom

grafnu:
Not sure why it's not showing up in influx, which is used as a test comparison... I'm missing something here I think -- looking for some hints.

### DIFF
--- a/faucet/gauge_pollers.py
+++ b/faucet/gauge_pollers.py
@@ -219,7 +219,7 @@ class GaugePortStatsPoller(GaugeThreadPoller):
             (('bytes', 'in'), stat.rx_bytes),
             (('dropped', 'out'), stat.tx_dropped),
             (('dropped', 'in'), stat.rx_dropped),
-            (('errors', 'out'), stat.tx_errors)),
+            (('errors', 'out'), stat.tx_errors),
             (('errors', 'in'), stat.rx_errors))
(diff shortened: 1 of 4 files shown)


gizmoguy:
Okay the output from the test made this more obvious, in FaucetUntaggedInfluxTest it calls _verify_influx_log() which does a double check that all the exported variables to influxdb exist, now you have added a new one you need to add it to the list:

```
diff --git a/tests/integration/mininet_tests.py b/tests/integration/mininet_tests.py
index ee1f1ab9..fc894b8a 100644
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -2005,8 +2005,8 @@ class FaucetUntaggedInfluxTest(FaucetUntaggedTest):
         self.assertTrue(os.path.exists(self.influx_log))
         expected_vars = {
             'dropped_in', 'dropped_out', 'bytes_out', 'flow_packet_count',
-            'errors_in', 'bytes_in', 'flow_byte_count', 'port_state_reason',
-            'packets_in', 'packets_out'}
+            'errors_in', 'errors_out', 'bytes_in', 'flow_byte_count',
+            'port_state_reason', 'packets_in', 'packets_out'}
 
         observed_vars = set()
         for _ in range(retries):
```

PR #3379, opened 2019-12-09
WIP: Don't edge_learn ports

grafnu:
My integration tests are running into a problem where sometimes they run into nasty forwarding loops -- I'm raising this PR just as a FYI and discussion point around the right fix. I'm still investigating the root cause and trying to create a reproducible test case... I *think* the actual problem is that sometimes learned flows don't stick and are rejected for some unknown reason -- and that combined with the edge_learn algorithm causes problems.

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -1209,8 +1209,8 @@ class Valve:
         Returns:
             list: OpenFlow messages, if any.
         """
-        #learn_port = self.flood_manager.edge_learn_port(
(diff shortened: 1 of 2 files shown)


anarkiwi:
Is there any more information please on what the issue is? The problem description is very vague, which makes commenting on this PR challenging.

If you don't have a specific problem to report yet, please close this PR and please log an issue with a reproducible bug when you are able.

PR #3397, opened 2019-12-19
Add metric for learned ports

grafnu:


anarkiwi:
Please add a test.

anarkiwi:
It also looks like associations will not age out (e.g if MAC moves ports or IP changes the old association will still be present).

grafnu:
Updated, PTAL. Test added, expiry added, and removed the l3 association since it's fundamentally l2 learning. Handles the moved case since there's only one port per dp/mac/vlan, so a move overwrites the previous value.

Issue #3356, opened 2019-11-21
new multidp tests cause fake hw tests to fail


gizmoguy:
We run a nightly job to run a full set of fake hardware tests on faucet master. The most recent test suite run failed due to #3343 

Here is how we run the tests:

```
docker build --no-cache --pull -t faucet/tests -f Dockerfile.tests .
docker run --rm --privileged --cap-add=NET_ADMIN \
       --sysctl net.ipv6.conf.all.disable_ipv6=0 \
       -v /var/local/lib/docker:/var/lib/docker \
       -v /tmp/faucet-pip-cache:/var/tmp/pip-cache \
       -e HWTESTS=1 \
       -e FAUCET_TESTS="" \
       -t faucet/tests
```

I would have expected the new tests to not run with these options but they seem to and they fail due to only one switch being available:

```
======================================================================
FAIL: test_tagged (mininet_multidp_tests.FaucetSingleStackStringOfDPTagged0Test)
All tagged hosts in stack topology can reach each other.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/faucet-src/tests/integration/mininet_multidp_tests.py", line 90, in test_tagged
    self.verify_stack_up()
  File "/faucet-src/clib/mininet_test_base_topo.py", line 446, in verify_stack_up
    self.fail('not enough links up: %f / %f' % (links_up, links))
AssertionError: not enough links up: 6.000000 / 8.000000

======================================================================
FAIL: test_tagged (mininet_multidp_tests.FaucetSingleStackStringOfDPTagged1Test)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/faucet-src/tests/integration/mininet_multidp_tests.py", line 104, in test_tagged
    self.verify_stack_up()
  File "/faucet-src/clib/mininet_test_base_topo.py", line 446, in verify_stack_up
    self.fail('not enough links up: %f / %f' % (links_up, links))
AssertionError: not enough links up: 6.000000 / 8.000000

======================================================================
FAIL: test_dp_one_vlan_from_router (mininet_multidp_tests.FaucetSingleUntaggedIPV4RoutingWithStackingTest)
Test when each DP contains a subset of the routed vlans
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/faucet-src/tests/integration/mininet_multidp_tests.py", line 831, in test_dp_one_vlan_from_router
    self.verify_stack_up()
[...]

PR #3343, opened 2019-11-15
Add Fault-Tolerance tests & generalise the Multi-DP integration tests

mab68:
- Add Fault-tolerance tests. Fault-tolerance tests can be run in the same manner as the integration tests but with the environment variable FAUCET_TESTS="-t". Fault-Tolerance tests will build a topology then randomly (or statically) tear down links or switches.

```
sudo docker run --privileged --rm --net=host --cap-add=NET_ADMIN \
    -e FAUCET_TESTS="-t" \
    -ti faucet/tests
```

- Add FaucetTopoTestBase, provides a generalised form of the integration test FaucetStringOfDPTest. Remove FaucetStringOfDPTest, all previous integration tests that inherited from the FaucetStringOfDPTest now inherit from FaucetTopoTestBase. These tests have also been moved to a new file.

- Add FaucetTopoGenerator, provides a generalised form of FaucetStringOfDPSwitchTopo. Remove FaucetStringOfDPSwitchTopo. Use FaucetTopoGenerator instead.

- Add TopologyWatcher. The fault-tolerance tests use the TopologyWatcher to determine what hosts should be connected etc.

FaucetTopoTestBase and the FaucetTopoGenerator allows for any Networkx graph (these can be made with the [Networkx graph generators](https://networkx.github.io/documentation/stable/reference/generators.html)) to be turned into a Faucet topology. 

```
dp_links = FaucetTopoGenerator.dp_links_networkx_graph(networkx.cycle_graph(n_dps))
host_links, host_vlans = FaucetTopoGenerator.untagged_vlan_hosts(n_dps, n_vlans)
self.build_net(n_dps=n_dps, n_vlans=n_vlans, dp_links=dp_links,
                      host_links=host_links, host_vlans=host_vlans)
```

PR #3398, opened 2019-12-19
Re-add fault-tolerance tests

mab68:
#3343


gizmoguy:
Note to self: double check that #3356 is fixed

gizmoguy:
LGTM, fixes #3356

PR #3409, opened 2020-01-09
Invalidate vlan caches on coldstart

grafnu:


anarkiwi:
You can use a write to stderr to do debugging. Thanks for adding the trace capability.

grafnu:
Updated to use stderr.

PR #3413, opened 2020-01-13
Need test to verify switches can match IP_DSCP field

brandonchau-github:
Locally ran test:
```
adding test FaucetDscpRewriteTest

Ran 0 tests in 0.000s

OK
running 0 tests in parallel and 1 tests serial
test_untagged (mininet_tests.FaucetDscpRewriteTest) ... *** Error setting resource limits. Mininet's performance may be affected.
ok

Ran 1 test in 29.534s

OK


test results
============
--- {}
...
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -897,8 +897,9 @@ Each acl contains a list of rules: a packet will have the first matching rule
 applied to it.
 
 Each rule is a dictionary containing the single key 'rule' with matches
-and actions. Matches are key/values based on the ryu RESTFul API. Actions
-is a dictionary of actions to apply upon match.
+and actions. Matches are key/values based on the `ryu RESTFul API.
+<https://ryu.readthedocs.io/en/latest/app/ofctl_rest.html#reference-description-of-match-and-actions>`_
+Actions is a dictionary of actions to apply upon match.
 
 .. list-table:: : acls: <acl name>: - rule: actions: {}
     :widths: 30 15 15 40
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -6755,6 +6755,76 @@ acls:
             source_host, overridden_host, rewrite_host, overridden_host)
 
 
+class FaucetDscpMatchTest(FaucetUntaggedTest):
+    # Match all packets with this IP_DSP and eth_type, based on the ryu API def
+    # e.g {"ip_dscp": 3, "eth_type": 2048}
+    # Note: the ip_dscp field is translated to nw_tos in OpenFlow 1.0:
+    # see https://tools.ietf.org/html/rfc2474#section-3
+    IP_DSCP_MATCH = 46
+    ETH_TYPE = 2048
+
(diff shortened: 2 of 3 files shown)


docs/configuration.rst
```diff
@@ -897,8 +897,9 @@ Each acl contains a list of rules: a packet will have the first matching rule
 applied to it.
 
 Each rule is a dictionary containing the single key 'rule' with matches
-and actions. Matches are key/values based on the ryu RESTFul API. Actions
-is a dictionary of actions to apply upon match.
+and actions. Matches are key/values based on the ryu RESTFul API.
+https://ryu.readthedocs.io/en/latest/app/ofctl_rest.html#reference-description-of-match-and-actions
```
gizmoguy: We could use an RST hyperlink here to make this a bit cleaner, e.g:

`ryu RESTFul API
<https://ryu.readthedocs.io/en/latest/app/ofctl_rest.html#reference-description-of-match-and-actions>`_

tests/integration/mininet_tests.py
```diff
@@ -6755,6 +6755,76 @@ def test_switching(self):
             source_host, overridden_host, rewrite_host, overridden_host)
 
 
+class FaucetDscpRewriteTest(FaucetUntaggedTest):
```
gizmoguy: This is technically not a `FaucetDscpRewriteTest` test, it's a `FaucetDscpMatchTest` or `FaucetDscpMatchEthDstRewriteTest`

gizmoguy:
Apart from those two minor changes LGTM. Push up a fix for both of those and I'll approve for merging.

docs/configuration.rst
brandonchau-github: Done!

tests/integration/mininet_tests.py
brandonchau-github: Done!

docs/configuration.rst
```diff
@@ -897,8 +897,9 @@ Each acl contains a list of rules: a packet will have the first matching rule
 applied to it.
 
 Each rule is a dictionary containing the single key 'rule' with matches
-and actions. Matches are key/values based on the ryu RESTFul API. Actions
-is a dictionary of actions to apply upon match.
+and actions. Matches are key/values based on the ryu RESTFul API.
+<https://ryu.readthedocs.io/en/latest/app/ofctl_rest.html#reference-description-of-match-and-actions>
```
gizmoguy: Syntax is still a little wrong here, you want to use the following exactly (RST is a bit weird for URLs):

```
`ryu RESTFul API
<https://ryu.readthedocs.io/en/latest/app/ofctl_rest.html#reference-description-of-match-and-actions>`_

Check out the embeddable URI documentation here: https://docutils.sourceforge.io/docs/user/rst/quickref.html#external-hyperlink-targets
```

brandonchau-github: My bad, updated this and did a manual inspection with the python previewer.  This should be corrected in the new patch.

PR #3426, opened 2020-01-15
Need test verifying set_field action for ICMPV4_TYPE, IPV4_DST, IPV4_…

brandonchau-github:
…SRC, IP_DSCP

Local test run:
```
adding test FaucetSetFieldsTest

Ran 0 tests in 0.000s

OK
running 0 tests in parallel and 3 tests serial
test_set_fields_generic_udp (mininet_tests.FaucetSetFieldsTest) ... *** Error setting resource limits. Mininet's performance may be affected.
ok
test_set_fields_icmp (mininet_tests.FaucetSetFieldsTest) ... ok
test_untagged (mininet_tests.FaucetSetFieldsTest) ... ok

Ran 3 tests in 82.950s

OK

```

### DIFF
--- a/clib/mininet_test_base.py
+++ b/clib/mininet_test_base.py
@@ -802,6 +802,15 @@ class FaucetTestBase(unittest.TestCase):
         return ('python3 -c \"from scapy.all import * ; sendp(%s, iface=\'%s\', count=%u)"' % (
             packet, iface, count))
 
+    def scapy_base_udp(self, mac, iface, src_ip, dst_ip, dport, sport, count=1, dst=None):
+        if dst is None:
+            dst = 'ff:ff:ff:ff:ff:ff'
+        return self.scapy_template(
+            ('Ether(dst=\'%s\', src=\'%s\', type=%u) / '
+             'IP(src=\'%s\', dst=\'%s\') / UDP(dport=%s,sport=%s) ' % (
+                dst, mac, IPV4_ETH, src_ip, dst_ip, dport, sport)),
+            iface, count)
+
     def scapy_dhcp(self, mac, iface, count=1, dst=None):
         if dst is None:
             dst = 'ff:ff:ff:ff:ff:ff'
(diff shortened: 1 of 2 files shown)


clib/mininet_test_base.py
```diff
@@ -802,6 +802,15 @@ def scapy_template(packet, iface, count=1):
         return ('python3 -c \"from scapy.all import * ; sendp(%s, iface=\'%s\', count=%u)"' % (
             packet, iface, count))
 
+    def scapy_base_udp(self, mac, iface, dport, sport, count=1, dst=None):
+        if dst is None:
+            dst = 'ff:ff:ff:ff:ff:ff'
+        return self.scapy_template(
+            ('Ether(dst=\'%s\', src=\'%s\', type=%u) / '
+             'IP(src=\'0.0.0.0\', dst=\'255.255.255.255\') / '
```
gizmoguy: An IP src of 0.0.0.0 and dst of 255.255.255.255 doesn't really make sense here, typically that is only used for DHCP packets (port 67/68). Can we add an argument to this function that allows us to pass in the ip src and dst of the test hosts? (you could set the IP DST to the local broadcast address for the subnet if you want the packet to go to all hosts).

clib/mininet_test_base.py
```diff
@@ -812,6 +821,15 @@ def scapy_dhcp(self, mac, iface, count=1, dst=None):
                  dst, mac, IPV4_ETH),
             iface, count)
 
+    def scapy_icmp(self, mac, iface, count=1, dst=None):
+        if dst is None:
+            dst = 'ff:ff:ff:ff:ff:ff'
+        return self.scapy_template(
+            ('Ether(dst=\'%s\', src=\'%s\', type=%u) / '
+             'IP(src=\'0.0.0.0\', dst=\'255.255.255.255\') / ICMP() / '
```
gizmoguy: An IP src of 0.0.0.0 and dst of 255.255.255.255 doesn't really make sense here, typically that is only used for DHCP packets (port 67/68). Can we add an argument to this function that allows us to pass in the ip src and dst of the test hosts? (you could set the IP DST to the local broadcast address for the subnet if you want the packet to go to all hosts).

clib/mininet_test_base.py
```diff
@@ -812,6 +821,15 @@ def scapy_dhcp(self, mac, iface, count=1, dst=None):
...
+             'BOOTP(op=1)') % (dst, mac, IPV4_ETH),
```
gizmoguy: A BOOTP payload isn't expected inside an ICMP packet, so remove it.

tests/integration/mininet_tests.py
```diff
@@ -6755,6 +6755,106 @@ def test_switching(self):
...
+class FaucetSetFieldsTest(FaucetUntaggedTest):
+    # A generic test to verify that a flow will set fields specified for
+    # matching packets
+    OUTPUT_MAC = '0f:00:12:23:48:03'
+    SRC_MAC = '0f:12:00:00:00:ff'
+
+    IP_DSCP_VAL = 46
+    # this is the converted DSCP value that is displayed
+    NW_TOS_VAL = 184
+    IPV4_SRC_VAL = "168.10.14.21"
```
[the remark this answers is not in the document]
brandonchau-github: Updated IPV4_SRC and DST vals to the recommended reserved one.

clib/mininet_test_base.py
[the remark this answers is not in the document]
brandonchau-github: Done!

clib/mininet_test_base.py
brandonchau-github: Updated this to use the test host IPs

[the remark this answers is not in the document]
brandonchau-github: Updated this to use test host IPs

PR #3435, opened 2020-01-21
Print host info when debugging multidp tests

mab68:


gizmoguy:
Could you de-stickler this one? Thanks!

Issue #3330, opened 2019-11-04
Make learning work when a LACP bundle is connected to multiple switches in a faucet stack


gizmoguy:
Commonly referred to by the term MC-LAG or MLAG, the ability to split LACP bundles across multiple distinct switches in the same faucet stack would be a very useful feature.

gizmoguy:
Further details available from @tweippert on faucet-users list:

https://lists.geant.org/sympa/arc/faucet-users/2019-09/msg00031.html
https://lists.geant.org/sympa/arc/faucet-users/2019-09/msg00034.html

gizmoguy:
Implementation detail: we will need to support sending a synthetic actor system ID when LACP links are spread across multiple faucet datapaths (at the moment we use dp_mac which is different per datapath).

gizmoguy:
This works now, if you set the ``lacp`` id to the same on multiple switches it will act as a single lacp bundle.

Note: currently if you override ``dp_mac`` and set your own custom mac addresses per switch then this could break our current mclag implementation. If anyone actually wants to do this let me know and we'll open a new ticket to fix that behaviour.

PR #3425, opened 2020-01-15
WIP: Remove actor_sync bit hardcoding

anurag6:
The idea is to set the Actor sync bit under certain conditions rather than hardcoding it to True. According to the logic below, we set it to 1 if lacp_forwarding is enabled on port(dp is stack root) or if there are no healthy stack roots. 

Testing: I am unsure about the testing that needs to be done. One test might be to check the dp member "stack_root_healthy" when there are no healthy roots available. Please advise.

PR #3447, opened 2020-02-03
Set sync, collecting & distributing only for LACP links on a chosen DP

mab68:
Nominate a DP for each LAG by the following conditions in order:

1. Number of UP (receiving LACP with sync bit set) LAG ports
2. Root DP
3. Lowest DPID

Only LACP links on the nominated DP will set sync, collecting & distributing
Prevent flooding to non-selected LACP links
Add LAG multidp tests


gizmoguy:
superseeds #3425

gizmoguy:
fixes #3330

PR #3459, opened 2020-02-10
Flood to stack ports for inter-VLAN routing

### LINKED ISSUE #3227: When multi DP routing on stacking, FAUCET router can't resolve hosts on non-root DPs (because route resolver doesn't flood to stack ports)
See also.

https://github.com/faucetsdn/faucet/pull/3226

mab68:
Fixes #3227


anarkiwi:
Looks like some pylint'ing needed?

PR #3478, opened 2020-02-27
images now have multi-arch for x86_64/arm

cglewis:
This is really a stop gap until `docker-compose-pi.yaml` can be fully removed when the remaining faucet/gauge/etc images push to docker hub with multi-arch.  This also removes the need for the Dockerfile.pi in https://github.com/faucetsdn/docker-prometheus.git

### DIFF
--- a/docker-compose-pi.yaml
+++ b/docker-compose-pi.yaml
@@ -2,7 +2,7 @@ version: '2'
 services:
     influxdb:
         restart: always
-        image: 'arm32v7/influxdb:1.7'
+        image: 'influxdb:1.7'
         ports:
             - '8086'
             - '8083'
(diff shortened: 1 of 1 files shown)


cglewis:
Didn't realize until after making this that the prometheus repo does additional stuff. if it makes sense to remove that and do separately let me know. that might need to be a build that does multi arch and then gets used rather than the official image

cglewis:
Looks like grafana is working out some kinks still: https://github.com/grafana/grafana/issues/23291

cglewis:
Updated with working versions of grafana, influxdb, and prometheus (which will match versions in the regular `docker-compose.yaml` so it's one step closer to being able to remove this file completely).  Additionally, this means that https://github.com/faucetsdn/docker-prometheus.git could be deleted/archived, since I think this is the only place it was being used.

aknuds1:
https://github.com/grafana/grafana/issues/23291 has been closed, and the issue should be taken care of :)

cglewis:
@aknuds1 thanks!

PR #3480, opened 2020-03-02
When stacking is enabled, only FAUCET MAC spoofing should only affect…

anarkiwi:
… non-stack ports.

Fixes https://github.com/faucetsdn/faucet/issues/3172


gizmoguy:
Thanks.

This caveat can now be removed from the tutorial as well:

https://github.com/faucetsdn/faucet/blame/master/docs/tutorials/routing.rst#L72-L75

PR #3489, opened 2020-03-12
Disable tcpdump user change

grafnu:


gizmoguy:
Hi @grafnu, I'm not sure how -Z works for tcpdump and just wanted to ask some questions.

If we don't specify -Z, is it true that tcpdump will still relinquish privileges down to some other user? If so, do you know what user it will use?

grafnu:
By default, if run as root, tcpdump will create the save file under the
user "tcpdump" -- essentially the default value is "-Z tcpdump""
"-Z root" does nothing if you're already root... since it changes user to
root.
-Z does nothing if you're not root.

All as per man page... they basically say "Set to -Z root to disable"...
kinda hacky, but that's what it's all about!

       -Z user
       --relinquish-privileges=user
              If  tcpdump  is running as root, after opening the capture
device or input savefile, change the user ID to
              user and the group ID to the primary group of user.

              This behavior is enabled by default (-Z tcpdump), and can be
disabled by -Z root.

I don't know what changed, but this started to be a problem for me only
recently...

gizmoguy:
Okay great, thanks for the explanation.

PR #3500, opened 2020-03-25
WIP: Prioritise selected link over standby

anurag6:
Change-Id: I5a78a9812e60202aee96d6b06afbc2fd014acabd

### DIFF
--- a/faucet/port.py
+++ b/faucet/port.py
@@ -530,10 +530,10 @@ class Port(Conf):
         """
         if self.lacp_collect_and_distribute:
             return 1, 1, 1
-        if self.is_port_standby():
-            return 1, 0, 0
         if self.is_port_selected():
             return 1, 1, 1
+        if self.is_port_standby():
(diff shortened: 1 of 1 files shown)


gizmoguy:
Hey @mab68, when you are free.

The intended behavior of this PR is that when a port is configured as `lacp_standby: True` we want to short circuit the state machine a little and have it end up with the lacp sync bit set, but not collect or distribute. So that when the backup link needs to take over it doesn't have to go sync --> C&D, it can just jump to C&D.

Is this PR going to do that? Or do we need to do more.

mab68:
@gizmoguy 

https://github.com/mab68/faucet/commit/845be8a3df4fdabb9c3abea93493aa23f0945424 should accomplish what you want.

gizmoguy:
Ah, great. That looks good to me, want to push it up as a PR that superseeds this one?

Thanks!

mab68:
superseeded by #3501

PR #3506, opened 2020-03-31
Add direct link to supported hardware

mubix:
This pull request just links directly to the line with supported hardware
to enable new users to find the list easier.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,7 @@ Hardware and software switch support
 
 Detailed guides for some switches are available on `readthedocs <http://docs.faucet.nz/en/latest/vendors/index.html>`_.
 
-FAUCET has been tested against the following switches (see also SUPPORTED_HARDWARE in `faucet/valve.py <faucet/valve.py>`_):
+FAUCET has been tested against the following switches (see also SUPPORTED_HARDWARE in `faucet/valve.py <faucet/valve.py#L2129>`_):
 
 - `Open vSwitch v2.1+ <http://www.openvswitch.org>`_
 - `Lagopus Openflow Switch <https://lagopus.github.io>`_
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,7 @@ Hardware and software switch support
 
 Detailed guides for some switches are available on `readthedocs <http://docs.faucet.nz/en/latest/vendors/index.html>`_.
 
-FAUCET has been tested against the following switches (see also SUPPORTED_HARDWARE in `faucet/valve.py <faucet/valve.py#L2129>`_):
+FAUCET has been tested against the following switches (see also SUPPORTED_HARDWARE in `faucet/valve.py <faucet/valve.py>`_):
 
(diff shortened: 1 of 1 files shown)


gizmoguy:
Hi @mubix thanks for your contribution.

There is a problem with linking directly to the line number in valve.py that this number will change frequently and we don't want to keep having to update it.

So I am wondering if there is a better way to fix this. Since the hardware list in README.rst is quite old too, I wonder if a better fix is just to delete that section and refer our users to docs.faucet.nz/en/latest/vendors/ which is the page we keep up to date.

The diff would look something like this:

```
diff --git a/README.rst b/README.rst
index cf913b3e..b17e9832 100644
--- a/README.rst
+++ b/README.rst
@@ -25,18 +25,7 @@ It supports:
 Hardware and software switch support
 ------------------------------------
 
-Detailed guides for some switches are available on `readthedocs <http://docs.faucet.nz/en/latest/vendors/index.html>`_.
-
-FAUCET has been tested against the following switches (see also SUPPORTED_HARDWARE in `faucet/valve.py <faucet/valve.py>`_):
-
-- `Open vSwitch v2.1+ <http://www.openvswitch.org>`_
-- `Lagopus Openflow Switch <https://lagopus.github.io>`_
-- Allied Telesis `x510 <https://www.alliedtelesis.com/products/x510-series>`_ and `x930 <https://www.alliedtelesis.com/products/x930-series>`_ series
-- `NoviFlow 1248 <http://noviflow.com/products/noviswitch>`_
-- Northbound Networks - `Zodiac FX <http://northboundnetworks.com/collections/zodiac-fx>`_ and `Zodiac GX <http://northboundnetworks.com/products/zodiac-gx>`_
-- Hewlett Packard Enterprise - `Aruba 5400R, 3810 and 2930F <http://www.arubanetworks.com/products/networking/switches/>`_
-- Netronome produces PCIe adaptors, via OVS - `Agilio CX 2x10GbE card <https://www.netronome.com/products/agilio-cx/>`_
-- Cisco Systems - `Catalyst 9000 Family Switches <http://www.cisco.com/c/en/us/products/switches/catalyst-9000.html>`_
+Detailed guides for supported switches are available on `readthedocs <http://docs.faucet.nz/en/latest/vendors/index.html>`_.
 
 Faucet's design principle is to be as hardware agnostic as possible and not require Table Type Patterns. This means that Faucet expects the hardware Open Flow Agent (OFA) to hide implementation details, including which tables are best for certain matches or whether there is special support for multicast - Faucet expects the OFA to leverage the right hardware transparently.
```

mubix:
Ya completely agree, and understood about the line numbers

gizmoguy:
Ah sorry, my patch might have been a little complicated to read due to formatting.

Feel free to delete the line that starts with "FAUCET has been tested against the following switches" and delete the list of switches (Open vSwitch/lagopus/etc).

gizmoguy:
LGTM. Thanks for making those changes for me.

PR #3538, opened 2020-04-28
Revert back to serialisable form of stack graph

anurag6:
The content of STACK_TOPO_CHANGE was changed recently as part of [fec52f](https://github.com/faucetsdn/faucet/commit/fec52f905b12046b95e32bdf5cc4fc859485ab31) However, the MultiGraph() object is not serializable and hence the STACK_TOPO_CHANGE event is emitted with 'graph' field empty. 

Currently, faucet tests dont check for the content of an emitted event. Please let me know if the tests are needed.

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -671,7 +671,7 @@ class Valve:
                     self.notify(
                         {'STACK_TOPO_CHANGE': {
                             'stack_root': valve.dp.stack_root_name,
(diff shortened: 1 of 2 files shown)


gizmoguy:
Hi @anurag6 since we already got this wrong in the past, if you could contribute a test that asserts STACK_TOPO_CHANGE is not generated with a None graph field that will ensure we don't do it again.

PR #3543, opened 2020-04-30
WIP: Consider lacp ports with NO_SYNC status to nominate active LACP link

anurag6:
Currently, Faucet does not set SYNC bit on backup links by default. So if the Active link goes down, the nomination process still selects the now down ACTIVE link and hence failover doesn't take place. 

This is because, currently in valve.get_lacp_dpid_nomination, we only consider dp.lags_up() or links with ACTIVE actor state while calculating the nomination. Expanding it to include links with NO_SYNC state ensures that should the ACTIVE link go down, the DP with a NO_SYNC backup link gets higher priority and gets nominated, ensuring failover.

TODO: tests for changes.

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -682,6 +682,10 @@ configuration.
         """Return ports that have LACP up."""
         return tuple([port for port in self.lacp_ports() if port.is_actor_up()])
 
+    def lacp_nosync_ports(self):
+        """Return ports that have LACP status NO SYNC."""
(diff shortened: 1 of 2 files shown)


anarkiwi:
This PR is pretty stale. Is there an still an intent to proceed with it?

anurag6:
This was done primarily as a demo to show intent regarding one of the changes. Let me close it for now.

PR #3551, opened 2020-05-05
Use a calculated combination of dp_id and port number for lacp port num instead of the port number

anurag6:
Currently, we set the lacp port id to the DP port number. However, this sends the same value on separate links if the same ports are configured. The PR combines the DP ID and the port number to generate a port ID that is used as the LACP port number instead. 

The PR has a test that tests for the LACP port num using the same algorithm.

### DIFF
--- a/faucet/port.py
+++ b/faucet/port.py
@@ -107,7 +107,7 @@ class Port(Conf):
         # Min time since last LACP response. Used to control rate of response for LACP
         'lacp_port_priority': 255,
         # Sets port priority value sent out in lacp packet
-        'lacp_port_id': 0,
+        'lacp_port_id': -1,
         # Sets port id value sent out in lacp packet
         'loop_protect': False,
         # if True, do simple (host/access port) loop protection on this port.
--- a/faucet/port.py
+++ b/faucet/port.py
@@ -107,6 +107,8 @@ class Port(Conf):
         # Min time since last LACP response. Used to control rate of response for LACP
         'lacp_port_priority': 255,
         # Sets port priority value sent out in lacp packet
+        'lacp_port_id': 0,
+        # Sets port id value sent out in lacp packet
         'loop_protect': False,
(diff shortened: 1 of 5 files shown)


faucet/valve.py
```diff
@@ -828,6 +828,7 @@ def ports_add(self, port_nums, cold_start=False, log_msg='up'):
                 continue
 
             if port.lacp:
+                port.dyn_lacp_port_id = ((self.dp.dp_id % 1000) * 100 + port.number) % 65535
```
grafnu, line 831: this does not at all seem like a sufficient range to avoid collisions.

anurag6: I used dp_id since it was readily available. Another way could be to use the last few digits of an md5 hash of the dp name. The suggested use of a list of dps makes sense but I odnt think that level of detail is exposed at the valve level.

grafnu:
md5 hash of the name doesn't solve the problem.  It's essentially a random
number, and so there's a chance of a collision when mapped down to % 1000
-- if you're going to probabilistic route, then likely need to calculate
the acceptable chance of a collision to determine the number of bits
required.

faucet/port.py
```diff
@@ -107,6 +107,8 @@ class Port(Conf):
         # Min time since last LACP response. Used to control rate of response for LACP
         'lacp_port_priority': 255,
         # Sets port priority value sent out in lacp packet
+        'lacp_port_id': 0,
```
gizmoguy, line 110: I think 0 could be a valid lacp_port_id value. Unsure if setting this to None will work with it being an `int` type in our config parser. Maybe -1 should be the default value that indicates faucet should autogen the value?

anurag6: Reset default value to -1.

anurag6:
I made one change to test for the most recent LAG_CHANGE event so as to get the role and state of an LAG that's up.

PR #3557, opened 2020-05-08
Add lacp port role to LAG_CHANGE event in addition to actor state

anurag6:
The PR helps track the lacp port role in addition to actor state in events. It modifies an existing check for the existence of the LAG_CHANGE event in the event logs to check for the presence of actor state and port role as well.


anurag6:
I made one change to test for the most recent LAG_CHANGE event so as to get the role and state of an LAG that's up.

PR #3570, opened 2020-05-26
Enable python 3.8 unit testing, and only install pylint/pytype when they are selected.

anarkiwi:


gizmoguy:
pytype just merged in python 3.8 support, will wait 1 week before merging this PR in the hopes they do a new release making this PR redundant.

anarkiwi:
Great! Though this PR does fix a 3.8 specific warning and it also doesn't install pylint or pytype if they don't have to be used.

gizmoguy:
Could you split the syntax fix into a separate PR?

Also in the ideal case it is probably best to run pytype on all python versions right?

anarkiwi:
OK I'll make a separate PR for the syntax fix.

Yes would be ideal I could make another PR to run it on all of them rather than just 3.6 as it is today. In theory though should give the same results given the same dependencies. Maybe there would be a difference, not sure.

PR #3591, opened 2020-06-03
Flush ports post datapath disconnect.

anurag6:
Currently, several port level structures for disconnected datapaths remain intact. Consequently, varz like port_lacp_state and port_status retain their previous values. The PR flushes the structures for disconnected datapaths.


anarkiwi:
port_flush() looks very close to a cut/paste of ports_delete() and will involve multiple calls to the same functions. If you want to explicitly down everything on a DP, you could ports_delete everything and just discard any resulting ofmsgs?

faucet/valve.py
```diff
@@ -798,6 +800,31 @@ def ports_delete(self, port_nums, log_msg='down', keep_cache=False, other_valves
...
+        Args: N/A
+        Returns: N/A
+        """
+        for port_num in self.dp.ports.keys():
+            if not self.dp.port_no_valid(port_num):
+                continue
+            port = self.dp.ports[port_num]
+            port.dyn_phys_up = False
+
+            self._set_port_status(port_num, False, time.time())
```
gizmoguy, line 815: Instead of generating time.time() here could you update _datapath_disconnect() in faucet.py to generate a `now` variable and pass this through to valve.datapath_disconnect().

This will make it similar to how _datapath_connect() works as well.

anurag6:
> port_flush() looks very close to a cut/paste of ports_delete() and will involve multiple calls to the same functions. If you want to explicitly down everything on a DP, you could ports_delete everything and just discard any resulting ofmsgs?

I created a new method, because I make a call to _set_port_status as well. I didn't want to do it in ports_delete because I did not wish to change its behaviour. If that is fine, I can go ahead and add that line to ports_delete and call ports_delete instead.

faucet/valve.py
anurag6: That makes sense. Made necessary changes.

anarkiwi:
Yes please please do change to use ports_delete - it will change its behavior but it sounds like you want to correct its behavior so that's absolutely better than adding duplicate code.

gizmoguy:
Sorry @anurag6 looks like this PR has a conflict with #3597 could you please rebase from master to resolve that conflict?

anurag6:
> Sorry @anurag6 looks like this PR has a conflict with #3597 could you please rebase from master to resolve that conflict?

Yup. Fixed the conflicts. Hopefully, I fixed most of the issues and the tests go through.

Issue #3509, opened 2020-04-08
Incorrect flood table sizes with stacking


mab68:
Faucet allocates the flood table sizes to 32

Switches 1-5 needs 44 flood table rules
Switches 6-12 needs 34 flood table rules

![topology](https://user-images.githubusercontent.com/33644913/78842722-f1281980-7a54-11ea-9b68-c83ae00b09a0.png)

CONFIG: [faucet.yaml.txt](https://github.com/faucetsdn/faucet/files/4453236/faucet.yaml.txt)

faucet-1 flowdump: 
[faucet-1 flowdump.txt](https://github.com/faucetsdn/faucet/files/4453277/faucet-1.flowdump.txt)

```
m@b68:/var/tmp$ cat 'faucet-1 flowdump' | grep table=3 | wc -l
44
```

Faucet-6 flowdump: 
[faucet-6 flowdump.txt](https://github.com/faucetsdn/faucet/files/4453283/faucet-6.flowdump.txt)

```
m@b68:/var/tmp$ cat 'faucet-6 flowdump' | grep table=3 | wc -l
34
```

Reference: #3198

mab68:
The correct formula for flood table sizes seems to be:
`(len(self.vlans) * (len(self.stack_ports) + 1) * 5) + 4`

mab68:
Fixed in #3597

PR #3597, opened 2020-06-03
Unit test reworks, FakeOFNetwork and generative unit tests.

mab68:
Add `FakeOFNetwork`; connects multiple `FakeOFTable`'s together. Traversal is tested by calling `network.is_output()` like `FakeOFTable` but also providing the source & destination DPIDs.

Change unit test `ValveTestSmall` to `ValveTestNetwork` and natively support the new `FakeOFNetwork`.

Add generative unit tests.
Run generative unit tests with: `-e FAUCET_TESTS="--generative_unit"`
- `ValveTopologyRestartTest`:
Test restarting onto different topologies.
- `ValveTopologyVLANTest`:
Test changing host `native_vlan` of a single host.
- `ValveTopologyTableTest`:
Test stacked flooding traversals between each host (using FakeOFNetwork).

Move tolerance tests to generative folder.
Tolerance tests are now run with: `-e FAUCET_TESTS="--generative_integration"`


mab68:
I should add that the generative tests pull from `networkx.generators.atlas.graph_atlas_g` which contains all graphs up to seven nodes. However, these are not multi-graphs so there are no redundant stack links being tested.

For now most of the generative unit tests will also fail because of #3509 .

mab68:
The following rule are currently being dropped on a host native VLAN change

`Priority: 20480 | Match:  eth_src 0e:00:00:00:00:01, in_port 3 | Instructions : drop`

PR #3604, opened 2020-06-09
Optimize out packet modifications that are not output or goto'd after…

anarkiwi:
…wards. Add test coverage to ensure optimization holds.


mab68:
This has my seal of approval.

PR #3633, opened 2020-07-22
Update lacp varz when LACP port role is updated

anurag6:
Currently, we only update varz and send out an event when the port state is changed. However, this leaves out corner cases during controller failover and in some cases, LACP link failover, where the role is updated internally but isn't reflected in the monitoring varz. 

The PR also updates test_nominated_dpid_port_selection to check for the varz after calling lacp_update instead of explicitly calling _reset_lacp_status.

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -833,7 +833,7 @@ class Valve:
         select_updated = self.switch_manager.lacp_update_port_selection_state(
             port, self, other_valves, cold_start=False)
         if updated or select_updated:
-            if updated:
+            if updated or select_updated:
                 self._reset_lacp_status(port)
             if port.is_port_selected() and port.is_actor_up():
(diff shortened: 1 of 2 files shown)


faucet/valve.py
```diff
@@ -833,7 +833,7 @@ def lacp_update(self, port, lacp_up, now=None, lacp_pkt=None, other_valves=None,
         select_updated = self.switch_manager.lacp_update_port_selection_state(
             port, self, other_valves, cold_start=False)
         if updated or select_updated:
-            if updated:
+            if updated or select_updated:
```
anarkiwi, line 836: looks like this 2nd if clause is now redundant and can be removed? line 835 does the same thing.

anurag6: Yup. Got overlooked. Thanks.

PR #3647, opened 2020-08-05
Refactor stacking into ValveStackManager & Stack classes

mab68:
`Stack`: All stack related DP methods are moved to a stack class `Stack` in `stack.py`. `Stack` provides the additional configuration option `down_time_multiple` that configures how many `FAUCET_STACK_ROOT_STATE_UPDATE_TIME` intervals for a down stack node to still be considered healthy.

`ValveStackManager`: All stack based algorithms in the various valve managers that don't directly belong in that valve manager have been moved to `ValveStackManager` in `valve_stack.py`. Part of the root nomination process has been moved to this class, along with the away & towards root port selection & pruning calculations. Tunnel updating has been moved to here as well.

`FAUCET_STACK_ROOT_STATE_UPDATE_TIME` environment variable configures the number of seconds to wait before checking stack root health. If the current root is unhealthy, a new root will be nominated.

### DIFF
--- a/clib/fakeoftable.py
+++ b/clib/fakeoftable.py
@@ -1020,7 +1020,7 @@ class FlowMod:
             name = 'output'
             if action.port == CONTROLLER_PORT:
                 value = 'CONTROLLER'
-            if action.port == IN_PORT:
+            elif action.port == IN_PORT:
                 value = 'IN_PORT'
             else:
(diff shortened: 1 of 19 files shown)


anarkiwi:
This is an amazing step forward - thank you! Will take a closer look ASAP.

PR #3649, opened 2020-08-06
Update to test-base 8.0.1 image.

gizmoguy:
* Installs mininet python package in venv.
* Adds Open vSwitch v2.13.1.
* Smaller image size.


gizmoguy:
@mab68 I managed to fix the fuzzer dockers, exabgp wasn't installing due to setuptools being too old.

PR #3656, opened 2020-08-10
Add unit test redundant ofmsgs offset check

mab68:
I had to generate the offset ofmsgs from the original (before reorganizing and optimizing) ofmsgs due to the prepared ofmsgs flow structure.


anarkiwi:
So cool!

PR #3692, opened 2020-09-20
Change root health & nomination calculation

mab68:
`min_stack_health` & `min_lacp_health` are just additional configuration options for deciding if a stack node is healthy.

We now keep track of the individual health percentages so deciding the best root candidate is easier. We choose a root by sorting a list of tuples taking favour of healthy nodes, followed by the highest health percentages, lastly followed by the priority.

### DIFF
--- a/clib/config_generator.py
+++ b/clib/config_generator.py
@@ -104,13 +104,33 @@ class FaucetTopoGenerator(Topo):
         """Return host port maps and link port maps"""
         return self._create_port_map(), self._create_host_port_map(), self._create_link_port_map()
 
-    def dp_dpid(self, i):
-        """DP DPID"""
+    def rand_sequential_dpid(self, i):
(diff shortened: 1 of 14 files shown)


grafnu:
I tried to run (a merged version of this) against our integration suite and there were some problems -- but I haven't had time to sort them out yet. So, I'd prefer not to merge this yet until I can suss out what's going on.  There are now some merge conflicts so I might have handled that incorrectly -- so maybe in the meanwhile if you could rebase this PR that might help.

grafnu:
I tried this out and seems to work fine!  We still have a problem with integration on our end, but that had to do with the stack change in 1.9.51 not this PR. So, from my perspective at this point I have no objections to this being merged (I didn't do any kind of code review on it, though -- just testing).  Thanks!

gizmoguy (approved):

PR #3695, opened 2020-09-25
self.stack_manager.chosen_towards_port might be None as stack is init…

anarkiwi:
…ializing with routing active


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/3694

gizmoguy (approved):

PR #3699, opened 2020-10-06
On a mirror change, we only need to update the mirrored port and flood rules - we don't have to delete the entire VLAN.

anarkiwi:


anarkiwi:
Fixes https://github.com/faucetsdn/faucet/issues/3698

gizmoguy (approved):

PR #3708, opened 2020-10-08
Clone stack and port state

grafnu:
Preserves the stack and port state through cold start.  I'm not sure what an appropriate test for this would be, pointers appreciated.


anarkiwi:
Sounds like a good test would be to use the unit test helpers to take port and stack state after bringing the network up, initiate a cold start, and compare with the copy before the test.

grafnu:
Closing for now until I add some tests.

PR #3713, opened 2020-10-11
Add test for changing stack root

grafnu:
Functionality neutral refactoring and adding a new test for changing stack root. The test has "wrong" values in it, which will be fixed in another PR.


gizmoguy:
superseded by #3716

PR #3718, opened 2020-10-12
Add RuntimeDirectory to faucet.service files

grafnu:


gizmoguy:
Just curious what /run/faucet is being used for here?

grafnu:
I think the actual directory is maps into is /var/run/faucet -- the
specific error that I was pointed out that this fixed was:

   - itamayo@zz-tri-esdn05-ctr2-1-1.n:~$ grep sock
   /var/log/faucet/faucet.log
   Oct 09 01:57:24 faucet ERROR    Unable to create event socket directory:
   [Errno 13] Permission denied: '/var/run/faucet'
   Oct 09 02:12:15 faucet ERROR    Unable to create event socket directory:
   [Errno 13] Permission denied: '/var/run/faucet'
   -

gizmoguy:
Right, I think where the confusion comes from is that FAUCET_EVENT_SOCK is disabled by default in faucet so with the default configuration you won't get this error.

The error only occurs when you set FAUCET_EVENT_SOCK to /var/run/faucet/event.sock or something similar which obviously points to a non-existent directory. I would have assumed the system that tells faucet to use this directory should create this directory (or you could use a path that does exist, e.g /var/run/faucet_event.sock).

If this is difficult then I guess in principle I would not be opposed to merging this, but it would create an empty directory (/run/faucet or /var/run/faucet depending on platform) on machines that don't use the event socket.

grafnu:
Ok -- we'll sort this out on our end and get back to you if there's
anything to follow-up on.

PR #3741, opened 2020-11-13
Add nosync links as a tie breaker if same number of or no up links

anurag6:


docker/runtests.sh
```diff
@@ -76,7 +76,9 @@ for opt in ${FAUCET_TESTS_SHORTENED}; do
             ;;
           u)
             # skip unit tests
-            UNITTESTS=0
+            UNITTESTS=1
+            DEPCHECK=0
```
gizmoguy, line 80: Should this set `INTEGRATIONTESTS` and `DEPCHECK` to 0 when `-u` is set?

Currently, `-u` is for skipping unit tests (rather than running only unit tests) so disabling code and integration checks when `-u` is specified at the moment seems wrong. I am open to reversing this logic if it makes more sense and make `-u` run only the unit tests, I'm happy to write that PR if you want that behaviour.

anurag6: I am sorry about that, my debug code made it into the PR while pushing the lint fixes. You are right, the file shouldnt be changed at all. Reverted it.

gizmoguy (approved):

PR #3855, opened 2021-06-28
Fix pylint errors on fuzz_config.py

lantz:


lantz:
This was one of the few tests that failed in my config, but upgrading pylint seemed to fix it, so it probably isn't necessary.

Issue #3865, opened 2021-07-07
Ports added via SIGHUP may stay link down


mwutzke:
When a port is added to the Faucet configuration via the SIGHUP process, and this then the port is added to the OpenFlow switch, the port will stay 'down' in Faucet, and no flows will be created, and no traffic will pass.

In order to see this issue, 2 items need to occur:
- a port needs to be added to a running Faucet configuration via a SIGHUP
- the port then needs to be added to an OpenFlow switch (eg `ovs-vsctl add-port ...`) that is already connected to a Faucet controller.

If any of these conditions do not exist, the setup seems to work. I suspect anything that results in Faucet processing a 'cold start' will not see the issue.

A quick overview of the reproduction is:
```
starting condition: Faucet is not running, and the OVS bridge does not exist

# 110
create YAML with 1 port defined (port 1)
start Faucet
# 120
update YAML with a 2nd port defined (port 1 and 2)
SIGHUP Faucet

# 130
create OVS bridge and connect to Faucet
create 2 link's in the UP state
attach these 2 link to the OVS bridge

verify: ping between hosts attached to the 2 links fails
```

The attached sighup.tgz should be able to reproduce the issue on a system which has an OVS installation, and which has Faucet installed (I used a `pip install` version for 1.9.53devXX).

Using these scripts, the following should show the issue:
```
cd sighup
./100-all
```

The fist `ping` sequence (`140-test`) will fail.

With this setup, the interesting sequence is when port2 get added (`130-ovs`):
```
INFO     DPID 100 (0x64) br100 status did not change: Port 2 up status True reason ADD state 0
INFO     DPID 100 (0x64) br100 Port 2 state OFPPS_LIVE reset, ignoring in expectation of port down
```

The equivalent (working) port1 logs are:
```
INFO     DPID 100 (0x64) br100 status change: Port 1 up status True reason ADD state 0
INFO     DPID 100 (0x64) br100 Port 1 (1) up
INFO     DPID 100 (0x64) br100 Configuring VLAN vlan100 vid:100 untagged: Port 1,Port 2
```

I suspect the issue is related to the OFPPS_LIVE log, that is only seen for port2 in this setup.

[sighup.tar.gz](https://github.com/faucetsdn/faucet/files/6775788/sighup.tar.gz)

mwutzke:
1. PR #3867 resolves the initial issue
2. This highlighted an inter-op issue with doveSnap
[...]

PR #3866, opened 2021-07-08
valve.py: don't assume port is up when adding it (#3865)

mwutzke:
When adding a port to a DP via a configuration update, the
port state is set to be 'up'. This results in future OFPPR_ADD
updates being ignored, and the port never getting into the
forwarding state.

This change leave new ports in the 'down' state, so that
correct OFPPR_ADD processing occurs.


mwutzke:
Abandoning in preference to a different way to resolve the issue

PR #3875, opened 2021-07-17
lint yaml files and add linter to test workflows

cglewis:


.github/workflows/tests-yaml-lint.yml
```diff
@@ -0,0 +1,11 @@
+name: Yaml Lint
```
gizmoguy, line 1: I think it's preferred to stylize YAML in upper-case rather than Yaml.

.github/workflows/tests-yaml-lint.yml
```diff
@@ -0,0 +1,11 @@
...
+on: [push, pull_request]
+jobs:
+  lintAllTheThings:
```
gizmoguy, line 4: Personal opinion here - but prefer job to be simply called yamllint

gizmoguy (changes requested):

.github/workflows/tests-yaml-lint.yml
cglewis: Changed.

.github/workflows/tests-yaml-lint.yml
cglewis: Changed.

gizmoguy (approved):

PR #3876, opened 2021-07-17
prevent potential for path traversal and pylint

cglewis:


lantz:
What is the justification for this? It seems to be removing useful functionality, and in particular the ability to not have a log?

cglewis:
@lantz the current implementation is susceptible to a path traversal attack like this one: https://security.openstack.org/guidelines/dg_using-file-paths.html

Currently all of the calls to this function are tests don't even use the argument to do anything other than always write out to a log file named `activate.log` so this seemed like a safer way to accomplish the same thing.

anarkiwi (approved):

PR #3884, opened 2021-07-21
Fix pre-commit and mypy issues

cglewis:


cglewis:
This only works for python3.6+

gizmoguy:
Suggest deferring this PR, as we may end up replacing the precommit hook with tox.

PR #3886, opened 2021-07-21
Miscalenous small cleanups/additions

mwutzke:
* A number of small usability / documentation improvements
* An initial Vagrant machine to isolate simulation from the developers host environment
* Force all scripts to explicitly use python3

### DIFF
--- a/tests/integration/mininet_tests.py
+++ b/tests/integration/mininet_tests.py
@@ -3819,7 +3819,7 @@ routers:
                 '10.0.0.0/24 next-hop 10.0.0.254',
                 '10.0.1.0/24 next-hop 10.0.0.1',
                 '10.0.2.0/24 next-hop 10.0.0.2',
-                '10.0.2.0/24 next-hop 10.0.0.2'):
+                '10.0.3.0/24 next-hop 10.0.0.2'):
             self.assertTrue(re.search(route_string, updates), msg=updates)
         # test nexthop expired when port goes down
         first_host = self.hosts_name_ordered()[0]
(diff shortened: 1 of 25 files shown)


tests/integration/mininet_tests.py
```diff
@@ -3819,7 +3819,7 @@ def test_untagged(self):
                 '10.0.0.0/24 next-hop 10.0.0.254',
                 '10.0.1.0/24 next-hop 10.0.0.1',
                 '10.0.2.0/24 next-hop 10.0.0.2',
-                '10.0.2.0/24 next-hop 10.0.0.2'):
+                '10.0.3.0/24 next-hop 10.0.0.2'):
```
cglewis, line 3822: Should the next hop be 10.0.0.3 here?

mwutzke: No I don't believe so, as the config used for this test-case has:
```
            - route:
                ip_dst: "10.0.3.0/24"
                ip_gw: "10.0.0.2"
```

cglewis: ah, good catch. I was just following the pattern and made an assumption that it should be 3.

anarkiwi:
Thanks for your contribution! This is quite a "busy" PR. Would you be able to help us by splitting it up into its various themes?

mwutzke:
Sure thing - can help with that. I think there are 2-3 I can split it into.

mwutzke:
Abandoning this uber-PR. Will break up into more discrete PRs

Issue #3853, opened 2021-06-22
Alpine >= 3.13 containers on old host OSs' need new libseccomp2


anarkiwi:
https://docs.linuxserver.io/faq#my-host-is-incompatible-with-images-based-on-ubuntu-focal-and-alpine-3-13

I just observed this trying FAUCET build on alpine 3.14, attempting to run on a Raspbian buster host. 

Not sure anything FAUCET can do about it other than warn users.

PR #3975, opened 2021-11-18
switch from ryu to os_ken.

anarkiwi:


anarkiwi:
Passes OK, but requires a chewie and beka release to allow for most recent eventlet/msgpack which os_ken provide.

gizmoguy:
New releases can be arranged for both of those, doing a chewie release involved building a new docker image so wondering if before we do a chewie release we upgrade the alpine used for the faucet/base & faucet/python3 images? Do we have a preferred alpine release? Looks like supported releases are v3.12 and v3.13 (both with python3.8, but v3.13 will trigger #3853) and v3.14 with python3.9.

anarkiwi:
> New releases can be arranged for both of those, doing a chewie release involved building a new docker image so wondering if before we do a chewie release we upgrade the alpine used for the faucet/base & faucet/python3 images? Do we have a preferred alpine release? Looks like supported releases are v3.12 and v3.13 (both with python3.8, but v3.13 will trigger #3853) and v3.14 with python3.9.

I propose v3.14, I've done extensive testing with this as a base.

gizmoguy:
Okay I'll work on getting 3.14 versions of our docker images out before the dev call tomorrow.

Btw alpine 3.15 just landed a couple days ago, which also uses python3.9 which we can look to upgrade to in future once we've been able to do some testing with it.

In the meantime, I've now packaged os-ken + eventlet + msgpack for apt: https://github.com/faucetsdn/python3-os-ken/runs/4361958898

gizmoguy (approved):

Issue #4104, opened 2022-06-14
sNAT via conntrack ACLs


courtland:
Hello,

I am hoping to get some some feedback with using Faucet + openvswitch as a basic IPv4 NAT router on an Ubuntu 22.04 system with untagged VLANs. After looking through some of your integration tests and the conntrack/ovs docs, I managed to put the following together. It seems to work, albeit the throughput performance is not great. It's not the most powerful hardware, but without conntrack in play I get line rate of about 1Gb/s routing between VLANs using Faucet. With conntrack sNAT it drops to 200Gb/s. I'm wondering if I am doing something wrong or if you have any ACL/flow optimization suggestions. Is there anything inherently wrong with NATing to a Faucet VIP?

```yaml
vlans:
  wan:
    vid: 4095
    description: "wan native"
    faucet_mac: "0e:00:00:00:10:02"
    routes:
      - route:
          ip_dst: "0.0.0.0/0"
          ip_gw: "10.8.8.1"
    faucet_vips:
      - 10.8.8.254/24
  lan:
    vid: 1
    description: "lan native"
    faucet_mac: "0e:00:00:00:10:01"
    faucet_vips:
      - 10.0.1.254/24
routers:
  vlan-router:
    vlans: [lan, wan]
acls:
  nat_out:
    - rule:
        eth_type: 0x0800 # ipv4
        ct_state: 0/0x20 # untracked
        actions:
          ct:
            flags: 1
            table: 1
            zone: 1 # unique snat zone
            nat:
              flags: 1
              range_ipv4_min: 10.8.8.254
              range_ipv4_max: 10.8.8.254
    - rule:
        actions:
          allow: 1
  nat_in:
    - rule:
        eth_type: 0x0800 # ipv4
        ct_state: 0/0x20 # untracked
        actions:
          ct:
            flags: 1
            table: 1
            zone: 1 # unique snat zone
            nat:
              flags: 1
    - rule:
        actions:
          allow: 1
dps:
  dp1:
    dp_id: 0x1
    hardware: "Open vSwitch"
    interfaces:
      2:
        name: "eno2"
        description: "eno2"
        native_vlan: wan
        acl_in: nat_in
      3:
        name: "eno3"
        description: "eno3"
        native_vlan: lan
        acl_in: nat_out

```
```shell
 ovs-vsctl show
c543bc62-82f3-416d-9e8e-3af4642499a4
    Bridge dp1
        Controller "tcp:127.0.0.1:6653"
[...]

PR #4132, opened 2022-08-05
Add a Connection Tracking Tutorial

courtland:
This is an attempt at introducing a connection tracking tutorial with stateful ACL rules and sNAT examples. Initially discussed in issue #4104 and via email with @gizmoguy.

The first two sections (stateful fw rules and sNAT) are complete, but the intended ICMPv6/ARP filtering section is still a TODO. I ran out of time for the time being, so I'm looking for some feedback on the tutorial thus far.

Thanks!

### DIFF
--- a/docs/tutorials/conntrack.rst
+++ b/docs/tutorials/conntrack.rst
@@ -1,0 +1,364 @@
+Connection tracking tutorial
+============================
+
+This tutorial will cover the use of the OVS connection tracking system (aka `conntrack`)
+in conjunction with Faucet.
+
+We will explore using the conntrack module to implement:
+
+    - Stateful firewall rules via conntrack ACLs
+    - Source Network Address Translation (sNAT)
+
+Prerequisites
+^^^^^^^^^^^^^
+
+- A basic understanding of OVS connection tracking concepts. The `OVS Conntrack Tutorial
+  <https://docs.openvswitch.org/en/latest/tutorials/ovs-conntrack/>`_ is a good starting
+  point.
+- A good understanding of the previous tutorial topics
+  (:doc:`acls`, :doc:`vlans`, :doc:`routing`)
+- Install Faucet - :ref:`tutorial-package-installation` steps 1 & 2
+- Install Open vSwitch - :ref:`tutorial-first-datapath-connection` steps 1 & 2
+
+- Install the `conntrack command line utility
+  <https://manpages.debian.org/testing/conntrack/conntrack.8.en.html>`_
+
+    .. code:: console
+
+        sudo apt-get install conntrack
+
+- Useful Bash Functions - Copy and paste the following definitions into your
+  bash terminal, or to make them persistent between sessions add them to the
+  bottom of your .bashrc and run 'source .bashrc'.
+
+    .. literalinclude:: ../_static/tutorial/as_ns
+       :language: bash
+
+    .. literalinclude::  ../_static/tutorial/create_ns
+       :language: bash
+
+    .. literalinclude:: ../_static/tutorial/cleanup
+       :language: bash
+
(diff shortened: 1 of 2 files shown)


docs/tutorials/conntrack.rst
```diff
@@ -0,0 +1,353 @@
...
+
+Our ping from host1 to host2 should continue to work, establishing an entry in the
+connection tracker. This time, however, host1's source IP of 10.0.0.1 gets NATed to the
+faucet VIP of 10.0.0.254.
+
+.. code:: console
+
+    as_ns host1 ping 10.0.1.2
+
+    tcpdump -n -e -ttt -i veth-host2 10.0.0.254
```
gizmoguy, line 347: Missing `host` from the tcpdump filter expression, this command should be:

`tcpdump -n -e -ttt -i veth-host2 "host 10.0.0.254"`

docs/tutorials/conntrack.rst
```diff
@@ -0,0 +1,353 @@
...
+
+We can debug how OVS interfaces with the conntrack module to deal with the tracked
+packet(s):
+
+.. code:: console
+
+    ovs-appctl ofproto/trace br0 in_port=1,tcp,nw_src=10.0.0.1,nw_dst=10.0.1.2
+
+Our ping from host1 to host2 should continue to work, establishing an entry in the
+connection tracker.
```
gizmoguy, line 226: Suggest adding a call to the `conntrack` command after the ping to demonstrate the entry has been created, the output would look like this:

```
$ sudo conntrack -L | grep 10.0.1.2
conntrack v1.4.5 (conntrack-tools): 19 flow entries have been shown.
icmp     1 26 src=10.0.0.1 dst=10.0.1.2 type=8 code=0 id=11240 src=10.0.1.2 dst=10.0.0.1 type=0 code=0 id=11240 mark=0 zone=10 use=
```

It would also need to be necessary to add a prerequisite to the top section of the tutorial for the user to install the conntrack as it's not a default (e.g. `apt-get install conntrack`)

docs/tutorials/conntrack.rst
```diff
@@ -0,0 +1,353 @@
...
+faucet VIP of 10.0.0.254.
+
+.. code:: console
+
+    as_ns host1 ping 10.0.1.2
+
+    tcpdump -n -e -ttt -i veth-host2 10.0.0.254
+
+
+ICMPv6 and ARP filtering
```
gizmoguy, line 350: Remove this section and it can be added in a subsequent PR.

gizmoguy (changes requested):
Great work, thanks for this @courtland!

I have just followed all the steps and confirm the tutorial is good. There are a couple minor things to change which I've added as comments.

If you remove the `ICMPv6 and ARP filtering` section from this PR, I can merge this one after the minor feedback has been addressed and I'm happy to write the `ICMPv6 and ARP filtering` section and add it as a separate PR.

docs/tutorials/conntrack.rst
courtland: Oops, thanks, I forgot to fix that after going through it all...

docs/tutorials/conntrack.rst
courtland: That's a good idea for now. I was thinking you have a better understanding on what to express here anyways.

courtland:
Thanks for the feedback, I believe I addressed everything. Let me know if I need to squash on my end.

gizmoguy (approved):
LGTM :+1: 

Squash down to a single commit and I'll merge it. Thanks!

courtland:
Squashed. Thanks!

gizmoguy (approved):

PR #4194, opened 2022-10-25
Update Grafana Dashboards to work with newer versions

courtland:
I discovered that the Grafana Dashboards offered in the docs no longer work with later versions of Grafana. Specifically 9.x in my case. For whatever reason, Grafana does not migrate them correctly. Importing the old dashboards works, but editing the panels results in an exception in the Grafana frontend.

I tried importing the dashboards in each of the latest major Grafana versions since 5.x (6->7->8->9). It seemed to attempt to migrate the dashboards, but most of the panels were still broken.

I believe to have narrowed this down to a few different problems that had to be tweaked manually.
1. The "table" and "graph" panel visualizations were deprecated in favor of a newer "table" panel, and the "timeseries" graphs. 
2. The variable regex syntax changed in the query `expr` format.

This PR attempts to update the default dashboards.

Additionally, I've removed the conversion of OF stats to rates within the default prometheus config, which is reflected in the new graph expressions. Since [Grafana 7.2+ has the new `$__rate` interval](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/), it seems better to let Grafana calculate the rates appropriately. Maybe I'm missing something?

I believe these Dashboards are mostly equivalent to the old ones, but if you all have a better way of "updating" them, I won't be offended if you want to scrap this PR and fix it in a different way. :)

### DIFF
--- a/docs/_static/grafana-dashboards/faucet_instrumentation.json
+++ b/docs/_static/grafana-dashboards/faucet_instrumentation.json
@@ -9,47 +9,62 @@
       "pluginName": "Prometheus"
     }
   ],
+  "__elements": {},
   "__requires": [
     {
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "5.3.4"
-    },
-    {
-      "type": "panel",
-      "id": "graph",
-      "name": "Graph",
-      "version": "5.0.0"
+      "version": "9.2.2"
     },
     {
       "type": "datasource",
       "id": "prometheus",
       "name": "Prometheus",
-      "version": "5.0.0"
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "timeseries",
+      "name": "Time series",
+      "version": ""
     }
   ],
   "annotations": {
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": true,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
         "type": "dashboard"
       }
     ]
   },
   "editable": true,
-  "gnetId": null,
+  "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
   "id": null,
   "links": [],
+  "liveNow": false,
   "panels": [
     {
       "collapsed": false,
+      "datasource": {
+        "type": "prometheus",
+        "uid": "000000001"
+      },
       "gridPos": {
         "h": 1,
         "w": 24,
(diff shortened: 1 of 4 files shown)


gizmoguy:
Hi @courtland 

Thanks for updating the grafana dashboards, something I had been meaning to do but hadn't got around to yet.

One interesting, side-effect of your change to using $__rate_interval is that this breaks older versions of prometheus (pre v2.22.0, see https://github.com/grafana/grafana/issues/44251).

That is quite an old version of prometheus, but unfortunately our "installing faucet for the first time" tutorial series, uses the ubuntu packaged version of prometheus (which for ubuntu focal is older than v2.22.0) rather than a binary install from https://prometheus.io/.

So how about you remove the $__rate_interval change from this PR, so that I can merge in the updates that make grafana v9 work, and I can try to come up with a way for the tutorials to not use the apt packaged version of prometheus.

courtland:
That's an unfortunate side-effect, I hadn't run into that. I'm running official prometheus and grafana docker images via podman.

I've replaced `$__rate_interval` with `1m` in my PR and squashed it. The dashboards still work for me without the need for the extra prometheus config.

The prometheus version in Jammy is more up-to-date than Focal. I also noticed that the prometheus snap released by Canonical is a bit more up-to-date, but I don't think that's necessarily the right way to go. I think it's good that the Faucet APT repo targets support for Focal, but since Jammy is the latest LTS, shouldn't the tutorial itself target that? Maybe in the tutorial: _This tutorial was written for Ubuntu 16.04_ , could just be changed to _This tutorial targets Ubuntu 22.04 and later_? I know not everyone is on latest LTS, but FWIW I haven't had any problems with the tutorials on 22.04 yet, other than this prometheus version thing.

Would you be open to changing the tutorial to run the lastest prometheus and grafana images via a simple docker-compose setup?

gizmoguy (approved):

gizmoguy:
Great, that is all working for me now, so merged.

I too briefly evaluated the prometheus snap, but the prometheus community didn't seem too happy with it (plus it wouldn't work on Debian where Buster has an old version as well).

One experience we had when presenting the tutorial live at conferences, is that people in the crowd had a wide mix of different versions installed on their laptop and it was always good to try and support as many of them as we reasonably could. I guess one downside of the apt approach is that it doesn't work on non-debian derived distros, such as fedora or arch, and the Docker approach would work well there.

What I'm thinking is that perhaps we should introduce a second installing faucet tutorial which uses Docker and people can pick which one they want to go through. We already have a [docker-compose.yml](https://github.com/faucetsdn/faucet/blob/master/docker-compose.yaml) which should spin up most of the right pieces, I guess the only downside is the tutorial would be quite a bit shorter as `docker-compose up` would do most of the work. What do you think about that idea?

courtland:
Yeah, I think using the existing docker-compose setup in a tutorial is a great idea. Especially if you're already using that for other testing/demonstration. At least for Grafana/Prometheus/etc you will have strong control over the versions used, and getting docker (or podman) going is relatively easy for most distros. Then you only have to worry about getting the OVS configuration correct.

Would you also suggest using the Docker image for Faucet/Gauge itself? I think in that case the build definition in the existing docker-compose.yml would need to be replaced with the latest image/tag.

PR #4256, opened 2023-03-07
Require eventlet v0.33.3 which works with modern dnspython.

gizmoguy:


cglewis (commented):
I think you still want dnspython, just not pinned to the lower version

gizmoguy:
Previous to https://github.com/faucetsdn/faucet/pull/4237 we didn't pin dnspython at all, as it's not a dependency of faucet itself, but of eventlet.

We only pinned the package here because eventlet weren't pinning it.

PR #4267, opened 2023-03-23
run black formatter on config_parser_util.py and valve_lldp.py

anarkiwi:


gizmoguy:
I do enjoy all the white space formatting changes, I think it makes parts of the code easier to read.

The one thing I'm wondering about is all the quotation mark changes. Here is the black documentation on why they picked double quotes for everything:

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

They mention the `--skip-string-normalization` option, if people aren't too fussed on single / double quotes, I wonder if we set this option to skip changing them all? Thoughts?

PR #4292, opened 2023-04-25
Upgrade to test-base 17.0.0.

gizmoguy:


cglewis (approved):
Latest will point to 23.04, yeah? I haven't even tried that yet...

gizmoguy:
Nah ubuntu-latest on github actions still points at 22.04. I think they only track LTS releases.

Getting an error on github actions still: `with return code 81: cgget: libcgroup initialization failed: Cgroup is not mounted\t\t",` will double check if there's some extra initialisation we need to do for cgroups in the github actions runner environment.

gizmoguy:
Closing this PR as a second one is needed to upgrade base image to Debian Bookworm.

PR #4443, opened 2024-01-14
Add Default Port Hint

flogriesser:
Added documentation about the default listening port (6653) for Faucet and how it can be customized in configuration files such as /etc/default/faucet, /etc/default/gauge, and /etc/faucet/ryu.conf. 

This information helps users easily adapt Faucet's port settings to their specific network requirements.

Feel free to further adapt this is just a suggestion. But at least for me this information would have been quite useful.
I did not know that I also have to adapt /etc/default/gauge that's why I ran into some issues

PS: Nice tool thanks a lot for the good work


gizmoguy (approved):

gizmoguy:
@flogriesser thanks for the patch!

PR #4733, opened 2025-11-10
VLAN ACL change detection to be able to warm reload

hieunt79:
I want to perform a warm reload when the VLAN ACL changes, so I made the following updates:
- For port changes:
    - Handle ports in changed_vlans and added_vlans separately.
    - When a port ACL changes (changed_acl_ports), handle it later through Valve._apply_config_changes().
- For VLAN changes: add: added_vlans, changed_acl_vlans
- Lastly: optimize return OpenFlow messages in Valve _apply_config_change()

### DIFF
--- a/faucet/valve.py
+++ b/faucet/valve.py
@@ -1580,7 +1580,9 @@ class Valve:
                 added_ports (set): added port numbers.
                 changed_acl_ports (set): changed ACL only port numbers.
                 deleted_vids (set): deleted VLAN IDs.
-                changed_vids (set): changed/added VLAN IDs.
+                changed_vids (set): changed VLAN IDs.
+                added_vids (set): added VLAN IDs.
+                changed_acl_vlans (set): changed/added VLAN ACL VLAN IDs.
                 all_ports_changed (bool): True if all ports changed.
                 all_meters_changed (bool): True if all meters changed.
                 deleted_meters: (set): deleted meter numbers.
@@ -1599,6 +1601,8 @@ class Valve:
             changed_acl_ports,
             deleted_vids,
             changed_vids,
+            added_vids,
+            changed_acl_vlans,
             all_ports_changed,
             _,
             deleted_meters,
@@ -1607,7 +1611,35 @@ class Valve:
         ) = changes
         restart_type = "cold"
         ofmsgs = []
-
+        self.logger.debug("""Valve changes are:
+            deleted_ports: {},
+            changed_ports: {},
+            added_ports: {},
+            changed_acl_ports: {},
+            deleted_vids: {},
+            changed_vids: {},
+            added_vids: {},
+            changed_acl_vlans: {},
+            all_ports_changed: {},
+            _: {},
+            deleted_meters: {},
+            added_meters: {},
+            changed_meters: {},
+            """.format(
+                deleted_ports,
+                changed_ports,
+                added_ports,
+                changed_acl_ports,
+                deleted_vids,
+                changed_vids,
+                added_vids,
+                changed_acl_vlans,
+                all_ports_changed,
+                _,
+                deleted_meters,
+                added_meters,
+                changed_meters,
+            ))
         # If pipeline or all ports changed, default to cold start.
         if self._pipeline_change(new_dp):
             self.dp_init(new_dp, valves)
@@ -1631,11 +1663,21 @@ class Valve:
 
         if deleted_ports:
             ofmsgs.extend(self.ports_delete(deleted_ports))
+
         if changed_ports:
             ofmsgs.extend(self.ports_delete(changed_ports))
+
         if deleted_vids:
             deleted_vlans = [self.dp.vlans[vid] for vid in deleted_vids]
             ofmsgs.extend(self.del_vlans(deleted_vlans))
+
+        # if vlan acl changed and there are changed/deletd vids then delete vlan acl.
+        if self.acl_manager and changed_acl_vlans.union(changed_vids, deleted_vids):
+            changed_vlans = [self.dp.vlans[vid] for vid in changed_acl_vlans.union(changed_vids, deleted_vids)]
+            for vlan in changed_vlans:
+                ofmsgs.extend(self.acl_manager.del_vlan(vlan))
+            self.logger.debug("Number of deleted Openflow messages generated: {}".format(len(ofmsgs)))
+
         # TODO: optimize for all meters being erased
         if changed_meters:
             # If a meter changed meter IDs, delete the old ID first and consider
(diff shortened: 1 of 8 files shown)


hieunt79:
Hi @gizmoguy, I've revisited this PR and fixed the failing test cases. Could you please review it?

faucet/valve.py
```diff
@@ -1631,11 +1661,21 @@ def _apply_config_changes(self, new_dp, changes, valves=None):
 
         if deleted_ports:
             ofmsgs.extend(self.ports_delete(deleted_ports))
+
```
gizmoguy: Remove whitespace here

faucet/valves_manager.py
```diff
@@ -345,9 +348,14 @@ def request_reload_configs(self, now, new_config_file, delete_dp=None):
         """Process a request to load config changes."""
         if self.config_watcher.content_changed(new_config_file):
             self.logger.info(
-                "configuration %s changed, analyzing differences", new_config_file
+                "configuration %s changed, start analyzing differences", new_config_file
```
gizmoguy: Let's keep the original text for this log message

gizmoguy:
Hi @hieunt79, I've been testing your patch and it seems to cover most scenarios and work well.

I've found one case that doesn't seem to work currently, and that is the case where all port ACLs are removed from a port, e.g start faucet with this configuration where we have 2 ports, both with ACLs:

```yaml
vlans:
    office:
        vid: 100
        description: "office network"
acls:
    block-ping:
        - rule:
            dl_type: 0x800      # IPv4
            ip_proto: 1         # ICMP
            actions:
                allow: False
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        interfaces:
            1:
                name: "host1"
                native_vlan: office
                acls_in:
                  - block-ping
            2:
                name: "host2"
                native_vlan: office
                acls_in:
                  - block-ping
```

and then remove the ACL from one of the ports and reload faucet, e.g:

```yaml
vlans:
    office:
        vid: 100
        description: "office network"
acls:
    block-ping:
        - rule:
            dl_type: 0x800      # IPv4
            ip_proto: 1         # ICMP
            actions:
                allow: False
dps:
    sw1:
        dp_id: 0x1
        hardware: "Open vSwitch"
        interfaces:
            1:
                name: "host1"
                native_vlan: office
                acls_in:
                  - block-ping
            2:
                name: "host2"
                native_vlan: office
```

We will see faucet do the warm reload and correctly detect the change in configuration:

```
Apr 22 21:50:49 faucet INFO     Reloading configuration
Apr 22 21:50:49 faucet INFO     configuration /etc/faucet/faucet.yaml changed, start analyzing differences
Apr 22 21:50:49 faucet INFO     Reconfiguring existing datapath DPID 1 (0x1)
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 no ACL changes
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 no VLAN changes
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 no METERS changes
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 no port changes
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 port 2 ACL changed (ACL ['port-block-ping'] to None)
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 ports where ACL only changed: {2}
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 Openflow messages generated: 2
Apr 22 21:50:49 faucet.valve INFO     DPID 1 (0x1) sw1 warm starting
```

However, the ACL flow rules do not get deleted from the datapath (unless you restart faucet to trigger a cold reload):

```
$ sudo ovs-ofctl -OOpenFlow13 dump-flows br0 | cut -d ',' -f 3- | sed -E 's/n_packets=[^,]+, //' | sed -E 's/n_bytes=[^,]+, //' | grep in_port=2 | grep table=0
 table=0, priority=20480,icmp,in_port=2 actions=drop
```

I also tested the same scenario for vlan ACLs and in that case everything worked fine, so it's just the port ACL case that seems to be broken currently.

hieunt79:
Hi @gizmoguy, thanks for pointing out the issue.

The generated flows are correct:

```
[OFPFlowMod(buffer_id=4294967295,command=3,cookie=11720004596349619985,cookie_mask=0,flags=0,hard_timeout=0,idle_timeout=0,instructions=[],match=OFPMatch(oxm_fields={'in_port': 3}),out_group=4294967295,out_port=4294967295,priority=20480,table_id=0), OFPFlowMod(buffer_id=4294967295,command=0,cookie=11720004596349619985,cookie_mask=0,flags=0,hard_timeout=0,idle_timeout=0,instructions=(OFPInstructionGotoTable(len=8,table_id=1,type=1),),match=OFPMatch(oxm_fields={'in_port': 3}),out_group=0,out_port=0,priority=20480,table_id=0)]
```

However, `remove_overlap_ofmsgs` in `valve_of.py` (lines 1269-1270) suppresses the delete flow:

```python
if deletes and addmod:
    by_kind["delete"] = remove_overlap_ofmsgs(deletes, addmod)
```

To fix this, I added a new `del_port_force` method to force the deletion of the port and any ACL rules belonging to it. Do you think this is a good approach?

I have also added a test case (FaucetConfigReloadPortAclRemoveTest) for this scenario. The latest commit passes both unit and integration tests.

gizmoguy:
Hi @hieunt79, thanks for digging into that edge case and adding a new test case, I understand better what is happening now.

I now see that to change an ACL on a VLAN or port, this patch will delete all the flow rules for the vlan/port and re-add the new flow rules from the new configuration. This approach is essentially what a cold-start does, except it does it for a whole datapath.

Here's what I am thinking is a better approach for implementing the warm start for ACL changes (which is similar to how we handle warm reloads for meter changes), that will truly be a warm start (i.e we only delete flow rules we are no longer using and add new flow that are missing):

1. In `_get_port_config_changes()` instead of returning the ports with changed acls in `changed_acl_ports`, for changed ports you could instead return either a list of all old port ACL objects (self.dp.ports.get(port_num).acls_in) and the new port ACL objects (new_dp.ports.get(port_num).acls_in), alternatively you could just return the added/removed/changed ACL objects
2. In `_apply_config_changes()` we can now loop over the port ACL objects and firstly call `acl_manager.del_port_acl()` on all old ACL objects, and then call `acl_manager.add_port_acl()` on new ACL objects. `remove_overlap_ofmsgs()` should hopefully cancel out the flowmods for ACLs that didn't change, but if not, it might be necessary to just delete the removed/changed ACL objects, and only add the new/changed ACL objects.

It should be possible to implement warm reload for VLAN ACLs in the same way, but I do note we are missing `acl_manager` methods for adding/deleting individual VLAN ACLs, so these would need to be added. I think that shouldn't be too difficult, but if it proves to be too hard then I guess the current implementation of deleting all the VLAN ACL flow rules and re-adding is currently working so we could probably merge the current implementation as is.

Let me know if you have any questions or if you find a problem in my design that I haven't considered that makes it difficult to implement or inefficient.

gizmoguy:
Hi @hieunt79, thanks for the hard work on this. I had a go at fixing some edge cases I found in this patch and submitted my own separate patch (#4819) which addresses these issues. I've credited your contribution on this issue by adding you as a co-author of the patch.

Please leave any comments and/or feedback on my PR if you find any issues with it and include any relevant faucet configuration files that are showing incorrect behaviour.

hieunt79:
Hi @gizmoguy, thanks for resolving this issue. Could you please correct my email in these commits? I made a typo, correct is nguyenhieu26496@gmail.com

gizmoguy:
@hieunt79 email should be updated now

PR #4800, opened 2026-05-06
Block-on-barrier: pause the controller for OFPBarrierReply

anarkiwi:
## Summary

- Add a per-datapath sender thread (`faucet/valve_send.py`) that owns the OF send path so the Faucet event loop never parks on `Event.wait`. The worker assigns the barrier xid itself, registers a `threading.Event` against `(dp_id, xid)` before `send_msg`, and blocks on the matching `EventOFPBarrierReply` (default 5s timeout, drops channel on miss).
- Wire `EventOFPBarrierReply` to a tiny handler in `faucet.py` that just `set()`s the waiter -- needed to keep the os-ken app loop responsive.
- Drop `@functools.lru_cache` on `valve_of.barrier()` -- the send path now mutates each request (xid, datapath), so the cached singleton was only safe by accident.
- Hook `_datapath_disconnect` to stop the sender and release pending waiters, instead of sleeping out the timeout.
- Re-enable `FaucetUntaggedApplyMeterTest`, `FaucetUntaggedMeterAddTest`, and `FaucetUntaggedMeterModTest` (the meter tests skipped pending this fix in #401).
- Doc rewrite of `docs/block_on_barrier.rst` to describe the as-built worker-thread design.

## Why a worker thread, not an in-loop wait

`os_ken.base.app_manager._event_loop` dispatches every event for an app serially on a single thread. A `send_flows` call that parked there waiting for the reply would deadlock against itself: the only code that can fire the wake is the same loop, and it can't run until the handler returns. The worker-thread design keeps the loop free.

## Test plan

- [x] Unit tests in `tests/unit/faucet/test_valve_send.py`: ordering, timeout, cancel-on-disconnect, two-datapath isolation, no-barrier passthrough.
- [x] Existing unit suite green (581 tests, ignoring the pre-existing `test_fctl::test_http_fail` flake which fails on `main` too -- it tries to talk to localhost:23).
- [x] black + pylint pass on changed files (pylint score for new module: 9.83).
- [ ] Integration: the three meter tests this PR re-enables run on the native hub.


anarkiwi:
Wrong base — retargeting to c65sdn/main.

PR #4799, opened 2026-05-01
Allow warm restart for VLAN/VIP, router, and ACL configuration changes

courtland:
## Summary

Changes to a VLAN's VIPs, router configuration, or ACL configuration currently force a cold restart — full datapath reconnection and a flow-table wipe. This is unnecessarily disruptive for deployments that update these dynamically (tenant provisioning, IPAM-driven VIP changes, dynamic router membership, ACL edits).

The warm-restart path (`del_vlan` + `dp_init` + `add_vlan`) already handles VLAN-level changes; two guards in `dp.py` were short-circuiting to cold, and ACL-only changes still went through `cold_start_port` which left stale flows on the wire. This PR removes the guards, fixes the cleanup, and replaces the ACL cold-start path with a diff-at-source granular reload that emits only changed rules — cost is O(k) where k = rules changed, regardless of total ACL size.

### What warm-starts now (previously cold)

| Scenario | Before | After |
|----------|--------|-------|
| Change VIP address on VLAN | Cold | **Warm** |
| Add/remove VIPs on existing VLAN (routing tables exist) | Cold | **Warm** |
| Change router VLAN membership (non-BGP) | Cold | **Warm** |
| Add/remove router (routing tables exist) | Cold | **Warm** |
| Add new VLAN with VIPs (OVS, routing tables exist) | Cold | **Warm** |
| Change a port's `acls_in` | Cold | **Warm** |
| Change a VLAN's `acls_in` / `acls_out` | Cold | **Warm** |
| Edit rules of an ACL referenced by a port or VLAN | Cold | **Warm** |

### What still cold-starts

- BGP config changes
- Router add/remove with no existing routing tables
- VID replacement when affected ports cover the DP
- Pipeline changes (e.g. first-ever VIP adds `ipv4_fib` / `vip` tables)
- TFM table-size changes (out of scope)
- All ports changed

### Scope rationale

VLAN/VIP and router warm-restart are inseparable: VIP changes must propagate to router-sibling VLANs for FIB consistency, and router membership changes drive the same VLAN reinstall plumbing. ACL warm-restart is technically a separable follow-up but rides the same `del_vlan / dp_init / add_vlan` infrastructure — absorbing it here (per offline thread with @gizmoguy) obsoletes #4733 in one merge instead of two.

[...]

PR #4807, opened 2026-05-19
Warm-reload for port and VLAN ACL configuration changes

courtland:
Warm-reload for changes to a port or VLAN's `acls_in` list.

This is a fresh PR following the closure of #4799, which combined this
work with VIP, router, and VLAN-base warm-restart. Per @gizmoguy's
feedback on scope and complexity, the new PR narrows to ACL only.

## Mechanism

Adopts the per-ACL design @gizmoguy proposed in his
[2026-05-01 review of #4733](https://github.com/faucetsdn/faucet/pull/4733#issuecomment-4357864093);
supersedes #4733 with @hieunt79 co-authored. For each port/VLAN whose
only change is `acls_in`, iterate the old and new ACL lists and call
`del/add` per-ACL with a cold-start priority decrement.
`remove_overlap_ofmsgs` cancels unchanged del+add pairs at flow-emit.

`add_port_acl` / `del_port_acl` gain @gizmoguy's suggested `priority`
kwarg (defaults to `auth_priority`, so dot1x callers unchanged).
`add_vlan_acl` / `del_vlan_acl` are the new VLAN-side methods
@gizmoguy noted were missing for per-ACL VLAN warm-reload.

## What was required to make the per-ACL approach work

- **`force_port_vlan` rules were landing in the wrong table.** The
  per-ACL install path hardcoded the same goto-instruction for both
  allow and `force_port_vlan` rules. Fine for dot1x (never uses
  `force_port_vlan: 1`) but wrong for arbitrary user ACLs. Fixed by
  deriving both instructions from the pipeline the way cold-start
  install does.

- **Stale flows when a port loses its last ACL.** Empty↔non-empty
  transitions need the wildcard rule flipped, which the per-ACL loop
  can't do, so they fall back to `cold_start_port`. That path had a
  latent bug @hieunt79 found in #4733: its flowdel and the wildcard
  flowmod shared a flowmodkey, so `remove_overlap_ofmsgs` cancelled
  the delete and old ACL flows stayed on the switch. Fixed by using
  a priority-less flowdel.

- **Old `acls_in` lists weren't available where the loop runs.**
  `dp_init` swaps `self.dp` partway through `_apply_config_changes`,
  so the per-ACL loop snapshots the old lists into local dicts before
  that point.

- **Priority decrement matches cold-start.** The loop's
  `priority -= len(acl.rules)` mirrors cold-start install's priority
  assignment so unchanged ACLs' del+add pairs have matching
  flowmodkeys and `remove_overlap_ofmsgs` can cancel them. That's how
  the optimization saves wire traffic.

- **Separate `changed_acl_vlans` bucket.** VLAN ACL changes used to
  fold into `changed_vlans` (forcing cold restart).
  `_get_vlan_config_changes` now signals them separately so they
  reach the warm-reload path.

## Tests

@hieunt79's two integration tests are included verbatim.
`FaucetConfigReloadPortAclRemoveAllTest` and a unit-level
`ValveRemoveAllPortACLsTestCase` cover the wildcard-cancellation case
fixed in the second bullet above. Adding or removing the
**first/last ACL on the DP** still cold-starts because the `port_acl`
table itself is conditionally allocated.

Closes #4733.

### DIFF
--- a/faucet/dp.py
+++ b/faucet/dp.py
@@ -1485,6 +1485,7 @@ class DP(Conf):
             changes (tuple) of:
                 deleted_vlans (set): deleted VLAN IDs.
                 changed_vlans (set): changed/added VLAN IDs.
+                changed_acl_vlans (dict): {vid: (old_acls, new_acls)}.
         """
         (
             _,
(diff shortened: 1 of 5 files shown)


gizmoguy:
This functionality has been implemented in #4815 and #4819.

PR #4826, opened 2026-07-12
Remove residual eventlet coupling; guard runtime stays eventlet-free

anarkiwi:
## Summary

Faucet already runs on os-ken's native (threading) hub (commit `eeab08f9`). This PR finishes the loose ends so **nothing in the runtime import graph reaches for eventlet**, and adds a regression guard.

Scoped to **runtime-free** (eventlet never imported/executed). eventlet remains an install-time transitive dep of `os_ken>=4.2.1`; removing it from site-packages entirely would require os_ken to make eventlet optional and is out of scope here.

## Changes

- **`requirements.txt`**: drop `beka==0.4.2`. Redundant with `c65beka==1.0.4` — both provide the `beka` namespace faucet imports via `from beka.beka import Beka` — and it was the one dragging in `eventlet>=0.33.3`. Which `beka` won previously depended on pip install order (latent bug). c65beka's `Beka` is API-compatible (verified constructor + `add_route`/`add_neighbor`/`run`/`shutdown`).
- **`tests/unit/faucet/test_no_eventlet.py`**: regression guard. Imports the runtime modules on the default hub in a clean subprocess and asserts `HUB_TYPE == native` and eventlet never entered `sys.modules`. Verified it has teeth (fails under `OSKEN_HUB_TYPE=eventlet`).
- **`valve.py` / `valves_manager.py`**: correct two comments that still asserted eventlet behavior (monkey patch / inotify conflict) that no longer happens on the native hub.

## Verification

Installed `requirements.txt` in a clean venv (upstream `beka` gone, only `c65beka` provides the namespace), then:
- new guard test: pass
- `test_main`, BGP config tests (`test_multi_bgp`, `test_default_bgp_connect`, `test_share_bgp_routing_vlan`): pass
- black + pylint clean on new/changed files


anarkiwi:
Opened against the wrong base repo (meant the c65sdn fork). Reopening there.

PR #4847, opened 2026-09-03
Replace os-ken with c65of

anarkiwi:
Depends on c65sdn/c65of#1. See `docs/drop_os_ken.rst` for the scope and
`docs/migration.rst` in c65of for the full list of behavioural differences.

## Why

os-ken sheds surface about once a major release and each shed has cost us a
fix: the eventlet demotion drove the block-on-barrier redesign, 4.0 deleted
`os_ken.cmd` and left 300 lines of `cmd/manager.py` inlined in
`faucet/__main__.py`, and an earlier release deleted `os_ken.app.wsgi` and
left 309 lines vendored in `ofctl_rest/`.

Every one of those breaks was in the runtime and framework layer, never the
wire layer. Of the 24,426 lines of os-ken our import closure touches, 3,155
account for 100% of the breakage.

Dropping it also drops oslo.config and eventlet.

## Verification

- The whole unit suite passes: 19 modules.
- **`tests/unit/faucet/test_handshake.py` is new**: it starts faucet as a
  process, connects a switch over TCP, and asserts the pipeline is
  programmed. No mininet, no OVS. Ten consecutive clean runs.
- A gap check over the tree resolved 325 distinct os-ken names against their
  c65of equivalents; none are missing.

The handshake test found two real bugs that 1,150 tests in c65of did not: the
datapath was not named before its features reply reached observers, so faucet
saw `unknown datapath DPID None` and dropped the channel; and a truncated
multipart reply surfaced as a bare `IndexError`. Both are fixed in c65of with
regression tests, the first verified to fail with the fix reverted.

## Operator-visible changes

None intended. The `--ryu-*` flags and `ryu.conf` are unchanged — they are
shipped in the systemd units — and the Prometheus metric labels for the echo
settings are unchanged. `ryu.conf` is now read with `configparser`, which
parses the same `[DEFAULT]` INI that oslo.config did.

## Two latent bugs this exposes

- `netaddr` is imported directly by five faucet modules and was never in
  `requirements.txt`; it only resolved because os-ken pulled it in. Now
  declared.
- `clib/valve_test_lib.py` read `icmp_pkt.type_` on an ICMPv4 packet, which
  raises `AttributeError` against os-ken. The rename to `.type` fixes it.

## Before merge

`c65of` is pinned by git URL. It has to be published to PyPI first: pbr
preserves a PEP 508 direct URL in the built metadata and PyPI refuses those,
so `c65faucet` would be unreleasable. Trusted publishing is provisioned; the
pin becomes a normal version once c65of is tagged.

https://claude.ai/code/session_01G39zbAFcoLMfAMkXpxn3sM


anarkiwi:
Opened against the wrong repository by mistake — this targets the c65sdn fork, not upstream. Closing; no review needed here. Apologies for the noise.

